PR #14, opened 2019-01-11
enhance the Dockerfile with default CMD to start

argan:


lovasoa:
@argan Thank you very much for your contribution, and sorry for the delay before merging it.

PR #16, opened 2019-07-17
Check name before passing to Filesystem

kimbtech:
I think it could be dangerous to pass url parts directly to the filesystem. Therefore i suggest to add some type of check. (e.g. make sure it is a valid ISO timestamp, as used in backup file names.)


lovasoa:
Thank you for your contribution, @kimbtech ! Simply out of curiosity: how did you find this issue ?

kimbtech:
I am hosting it on my own private instance and made some changes, when I just saw the issue. So it was just a coincidence.
By the way, I really like the tool.

PR #21, opened 2020-03-26
WIP: Add Background Tool, Pointer Tool, Minio Support

654wak654:
**When merged this pull request will:**
- Add a background tool (Shortcut: b): Allows you to upload an image for the background of the canvas.
  - [ ] Some setting to move/scale the background image?

- Add a pointer tool (Shortcut: f): Allows users to point at things for others to see.
  - [ ] Everyone should get their own pointer.
  - [ ] Showing pointers to new users should be done in a better way.

- Add support for [minio](min.io), object storage.
  - [ ] Currently it replaces all filesystem use, but filesystem use should be brought back, and minio made an optional thing.
  - [ ] Minio credentials should be read from an environment variable instead of js code itself.

**Other TODOs:**
- [ ] Should probably add the vscode and idea folders to .gitignore.
- [ ] Clean back the README
- [ ] Use IIFEs instead of js modules.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,2 @@
+# Default ignored files
+/workspace.xml
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+> Hi. This is a fork of the original project. It adds couple of small features (2 new tools and minio support). They are described briefly at the bottom of this README.
+
 # WBO
 
 WBO is an online collaborative whiteboard that allows many users to draw simultaneously on a large virtual board.
@@ -73,3 +75,12 @@ If you do that, the code is running directly on your machine, without any isolat
 ## Troubleshooting
 
 If you experience an issue or want to propose a new feature in WBO, please [open a github issue](https://github.com/lovasoa/whitebophir/issues/new).
+
+
+### Additional Features
+#### Two new tools:
+- Background (Shortcut: `b`): Allows you to upload an image for the background of the canvas. 
+- Pointer (Shortcut: `f`): Allows users to point at things for others to see.
+
+#### Minio suport:
+- Instead of writing to the filesystem, the server now saves boards to a minio instance. The configuration is inside  [server/minio.js](server/minio.js).
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-> Hi. This is a fork of the original project. It adds couple of small features (2 new tools and minio support). They are described briefly at the bottom of this README.
+# 654wak654 Fork
+Hi. This is a fork of the original project. It adds couple of small features (2 new tools and minio support). They are described briefly at the bottom of this README.
 
 # WBO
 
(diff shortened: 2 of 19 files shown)


README.md
```diff
@@ -73,3 +76,12 @@ If you do that, the code is running directly on your machine, without any isolat
...
 If you experience an issue or want to propose a new feature in WBO, please [open a github issue](https://github.com/lovasoa/whitebophir/issues/new).
+
+
+### Additional Features
+#### Two new tools:
+- Background (Shortcut: `b`): Allows you to upload an image for the background of the canvas. 
+- Pointer (Shortcut: `f`): Allows users to point at things for others to see.
+
+#### Minio suport:
+- Instead of writing to the filesystem, the server now saves boards to a minio instance. The configuration is inside  [server/minio.js](server/minio.js).
```
lovasoa: I don't think we want the changes to the README

.idea/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Default ignored files
+/workspace.xml
```
lovasoa: Maybe we should add .idea and .vscode to the gitignore instead of checking them in.

client-data/tools/background/background.js
```diff
@@ -0,0 +1,37 @@
...
+    fakeCanvas.style.background = `url("${fileURL}") 170px 0px no-repeat`;
+}
+
+Tools.add({
+    "name": "Background",
+    "icon": "🖼️",
+    "shortcut": "b",
+    "draw": draw,
+    "onstart": onstart
+});
```
lovasoa: I'm not sure about that tool. It could quickly get out of hands with many users...

client-data/tools/pointer/pointer.js
```diff
@@ -0,0 +1,93 @@
...
+        "press": () => highlightPointer(true, true),
+        "move": move,
+        "release": () => highlightPointer(false, true)
+    },
+    "draw": draw,
+    "onstart": () => showPointer(true, true),
+    "onquit": () => showPointer(false, true),
+    "mouseCursor": "none",
+    "stylesheet": "tools/pointer/pointer.css"
+});
```
lovasoa: This tool seems very useful ! This is the one I think we need the most. But it should have one pointer per user.

server/boardData.js
```diff
@@ -234,26 +225,43 @@ BoardData.prototype.validate = function validate(item, parent) {
...
+			dataStream.on("end", () => {
+				try {
+					boardData.board = Buffer.concat(buffer).toJSON(); // TODO: Might have to do string -> json first
+					for (id in boardData.board) boardData.validate(boardData.board[id]);
+					log('disk load', { 'board': boardData.name });
+				} catch (err) {
+					catchCallback(err, false);
+				}
+				accept(boardData);
+			});
```
lovasoa: Maybe we could have a simple Persistence interface, that just asynchronously writes or read a value for a key, and have the backend (file or minio) chosen using an environment variable.

README.md
654wak654: Yup, it's in my TODO list to just remove these.

.idea/.gitignore
654wak654: Yup, in the TODO list.

client-data/tools/background/background.js
654wak654: Yes it's not something for public boards. Maybe it could be enabled when creating a board? Or even in the server?

PR #24, opened 2020-04-03
Adding Color Presets

adrian1001:
Added a small color palette, accessible also via keyboard shortcuts 0..9.
Can be deactivated by setting an empty list of preset colors.

Use case:
When doing a joint whiteboard session in a team, it improves the visual result if everybody sticks to a small number of predefined colors rather than picking from the whole palette.

### DIFF
--- a/server/sockets.js
+++ b/server/sockets.js
@@ -51,6 +51,10 @@ function socketConnection(socket) {
 		});
 	}
 
+	socket.on("error", noFail(function onError(error) {
+		log("ERROR", error);
+	}));
+
 	socket.on("getboard", noFail(function onGetBoard(name) {
 		joinBoard(name).then(board => {
 			//Send all the board's data as soon as it's loaded


client-data/js/board.js
```diff
@@ -26,6 +26,15 @@
...
 
+Tools.i18n = (function i18n() {
+	var translations = JSON.parse(document.getElementById("translations").text);
+	return {
+		"t": function translate(s) {
+			return translations[s] || s;
+		}
+	};
+})();
+
```
adrian1001, line 37: Moved this up here, because I got a not-defined warning in Firefox for a line where Tools.i18n was referenced

lovasoa:
Hey @adrian1001  ! I wanted to rebase this PR against the latest master, but I accidentally pushed the wrong commit to your repo, which closed this PR, and removed my permission to push to your fork. Can you please reopen it ? I'll finish the rebase and we can then discuss your changes.

lovasoa:
About the PR itself: 

It looks very good to me, I'm ready to merge it. But I'd like to remove the use of non-essential modern javascript features. The code currently does not go through any compiler, and is compatible even with ancient browsers. I'd like to update the code a little to replace the `Map` with a simple array of arrays.

Also: the CSS changes seem to make the tools' tooltip opening animation look weird, can we fix that ?

lovasoa:
@adrian1001 , if you are okay with my changes in #25, I'll merge them.

PR #25, opened 2020-04-03
Color presets

lovasoa:
Followup on #24 by @adrian1001

### DIFF
--- a/client-data/board.css
+++ b/client-data/board.css
@@ -46,7 +46,7 @@ html, body, svg {
 	cursor:default;
 	padding: 10px;
 	pointer-events: none;
-	width: 300px;
+	width: max-content;
 }
 
 #menu.closed {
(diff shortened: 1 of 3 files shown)


adrian1001:
Hi Ophir,
what a quick response!

Tested #25, works as intended, suggest to merge it.

Thanks for fixing all the stuff I broke! ;-)  <3

- The Map was indeed so unnecessary.
- Didn't even notice that my change broke the animation, sorry
- Ah, that's why there were random colors
- Out of curiosity: What are you using for validating js?
- If you care: The (button.key !== undefined) that I introduced, could also be replaced by (button.key). Patch below.

And: Thanks for sharing that nice piece of software! Simple but very useful!

BR,
Adrian


```
diff --git a/client-data/js/board.js b/client-data/js/board.js
index 8659e0b..57663aa 100644
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -100,7 +100,7 @@ Tools.HTML = {
 			elem.addEventListener("click", setColor);
 			elem.id = "color_" + button.color.replace(/^#/, '');
 			elem.style.backgroundColor = button.color;
-			if (button.key !== undefined) {
+			if (button.key) {
 				elem.title = Tools.i18n.t("keyboard shortcut") + ": " + button.key;
 			}
 		});

```

lovasoa:
Add a list of predefined colors and shortcuts for them.

lovasoa:
Merged. Thank you very much for your contribution, @adrian1001 !

Issue #10, opened 2018-12-04
Possibility to edit / move entities


tracim:
Currently: if I want to modify an entity, I must delete the current one and create a new one. It's ok for basic entities, but as soon as my schema is made of several entities, it will be complicated.

Expected: allow to modify an entity. The minimum would be to be able to move an entity.

This would be a required feature for us to integrate the tool in tracim as a dedicated app

lovasoa:
This is a good feature idea!

gsutay:
Hi Ophir,

Can we collaborate on this? My goal is to complete this feature in 1 week.  Starting today, I can put about 40 hours into this effort. I am familiar with JavaScript and understand the WBO architecture diagram but I am new to WBO development and working with websockets/memory.

This is 2 features to me (edit and move). My objective is to be able to select and move board objects.

I am thinking we select objects now with mouse pointer for delete.  Starting with same object selection, maybe we can assign object to mouse pointer location (instead of deleting) until mouse clicked again. This would provide a select object > drag and drop result.

Let me know what you think.

Cheers,
Gene

lovasoa:
Hello @gsutay,
  It is great to see someone ready to work on this ! Of course, I am ready to help and give guidance. Do not hesitate to open new issues as you start working on the code. 

  More specifically, about the implementation you propose : 
  - Currently, the eraser works by listening to `mouseover` events on the SVG elements on the board. This makes it difficult to erase small object, and I have plans to change that : https://github.com/lovasoa/whitebophir/issues/8 . I don't think object selection should work this way.
 - An idea I had is that object selection could work by creating an SVG rectangle which would both give visual feedback about the selection, and get the list of selected elements using an [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

Do not hesitate to open a PR even before you have something working, so that we can discuss your code as soon as possible.

Thanks for your commitment !

P.S. By the way, out of curiosity, what are you using wbo for ?

gsutay:
@lovasoa

Thanks for the quick reply. I am reading up on the related technologies currently.

My use case would be for meetings, to replace a physical whiteboard with a virtual board where people can select board elements and then move them around (as is done in SAFe training sessions with sticky notes on a physical whiteboard).  Some of the commercial boards I have looked at have this feature already.

Will share once I have some progress on POC to show.

gsutay:
@lovasoa

[...]

PR #29, opened 2020-04-08
added selector tool to move objects

Mk-arc:
Hi there,

note: This is still work in progress. I just want to inform you as soon as possible that something is happening here.

I started implementing a "selector"-Tool to move objects after they have been placed. So far it works for Text, Lines and Rectangles... Pencil strokes are still up for implementation.

**Problems that i noticed so far:**

- Moving emits a lot of updates to the servers, which results in the client getting blocked when a "drag" is done for long time.
- I dont really like that the client is working directly on the svg datatypes, while the server is working on abstract datatypes. It would be nice when they work on the same elemets. However, so far, the client doesnt know anything about this objects.
- Moving pencil strokes is not simple...

Also... this is the first time I worked with javascript... Im coming more from the embedded C++ world. So probably a lot of stuff I did is garbage. It would be nice to receive some help on this issue so we can finish this features fast. Its important :)

### DIFF
--- a/client-data/board.html
+++ b/client-data/board.html
@@ -85,6 +85,7 @@
 	<script src="js/path-data-polyfill.js"></script>
 	<script src="js/minitpl.js"></script>
 	<script src="js/board.js"></script>
+	<script src="tools/selector/selector.js"></script>
 	<script src="tools/pencil/pencil.js"></script>
 	<script src="tools/line/line.js"></script>
 	<script src="tools/rect/rect.js"></script>
--- a/client-data/tools/selector/selector.js
+++ b/client-data/tools/selector/selector.js
@@ -1,0 +1,148 @@
+/**
+ *                        WHITEBOPHIR
+ *********************************************************
+ * @licstart  The following is the entire license notice for the
+ *  JavaScript code in this page.
+ *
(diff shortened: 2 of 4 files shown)


Mk-arc:
Ohhh I just saw that someone else is working on this issue #10

lovasoa:
Awesome, let's work on this together.

Just a question before I dive deeper into your code: why did you choose to create a new "modify" event type instead of reusing "update" ?

lovasoa:
>     * Moving emits a lot of updates to the servers, which results in the client getting blocked when a "drag" is done for long time.

I encountered this problem before, and the easiest solution is to throttle the updates. See how this is done in the pencil tool :  
https://github.com/lovasoa/whitebophir/blob/master/client-data/tools/pencil/pencil.js#L61-L67

If we need it in the selection tool too, then we may want to move the throttling code to board.js so that it can be reused by all tools.

Mk-arc:
> Awesome, let's work on this together. Just a question before I dive deeper into your code: why did you choose to create a new "modify" event type instead of reusing "update" ?

Fair question. I think theres no really specific reason. I guess it seemed more suitable for me to just "modify" the values that are actuallly considered by the change instead of overwriting the whole object.

PR #30, opened 2020-04-11
Do not install on run

jrd:
The docker image should contains already resolved dependencies.

No need for a start.sh, you could just run node directly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,4 +9,4 @@ EXPOSE 80
 
 VOLUME /opt/app/server-data
 
-CMD ["/usr/local/bin/node", "/opt/app/server/server.js"]
+CMD ["npm", "start"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,16 +1,12 @@
...
 
 VOLUME /opt/app/server-data
 
-RUN touch /usr/bin/start.sh # this is the script which will run on start
-
-RUN echo 'npm install --production' >> /usr/bin/start.sh
-RUN echo 'node /opt/app/server/server.js' >> /usr/bin/start.sh
-
-CMD ["/bin/sh","/usr/bin/start.sh"]
+CMD ["/usr/local/bin/node", "/opt/app/server/server.js"]
```
lovasoa, line 12:
```suggestion
CMD ["npm", "start"]
```

PR #34, opened 2020-04-22
Fix path used for board download

iblech:
Most places in the code reference the `server-data` directory as `__dirname/../server-data`. There is a single occurrence where it is referenced directly as `./server-data`.

This pull request harmonizes the usage, so that all references are indirect via `__dirname`.

(I noticed this when packaging WBO for NixOS. There we need to decouple the `server-data` directory from the directory where the source files are stored, as the latter directory is readonly in NixOS. In the future, this decoupling could be an option for WBO; meanwhile, a simple `sed` patch gets the job done.)

### DIFF
--- a/client-data/index.css
+++ b/client-data/index.css
@@ -45,7 +45,7 @@ main {
 }
 
 main>div {
-	margin:25px;
+	margin:15px;
 	min-width: 250px;
 	padding:35px;
 	transition:.07s;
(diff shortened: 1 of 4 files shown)


lovasoa:
Thank you for finding and reporting that @iblech !

I fixed it, and the persistence folder can now be configured with the `WBO_HISTORY_DIR` environment variable.

iblech:
Awesome, thank you for the proper fix! :-)

PR #36, opened 2020-04-22
Be more vigilant in validating file names

iblech:
See commit message; this is a non-exploitable security issue. (If it was exploitable, I would follow a more responsible disclosure process.)


lovasoa:
OK, thank you !

Issue #31, opened 2020-04-15
Not really an issue but...


rdbeach:
Hey. Thanks for a great board. I have forked this board and made some alterations (actually quite a bit), mostly for my own personal use as a math tutor. Check them out at [rdbeach/wb](https://github.com/rdbeach/wb) So far, the changes have undergone limited testing on Safari and Chrome, both on my laptop and on my phone. Seems to work pretty good so far.

A list of changes:

Redesign of the GUI/layout
Added whiteout tool
Added circle tool
Modified text tool so that the input appears closer to the text
Made the removal tool more robust so you can remove small objects easier 
Added an optional grid
Added an eraser tool
Added a clear board tool
Added an undo tool
Added a redo tool
Added a zoom in and zoom out tool in separate files.
Made some changes to the server to accommodate the new tools.

lovasoa:
Your work is great ! Congratulations @rdbeach !

I am reposting here [the comment I made on your commit](https://github.com/rdbeach/wb/commit/b532b7b8fec9348f7373430882fc5b8138833ed3#commitcomment-38503451) :

> Hello @rdbeach !
> I stumbled on this commit on my github feed. It looks like you've done some great work on WBO ! 👍 Would you be interested by working with me on integrating some of these changes back into WBO ? This would give them more visibility, reduce the maintenance burden on your fork, and potentially enable some more contributions by other parties. I am especially interested by your new eraser tool. Working with svg masks is a great idea that I wish I had from the beginning !
There are some browser compatibility and other issues with your changes, though, and I would be glad to work with you on solving these.
> Cheers,
> @lovasoa  

Also, I think you forgot to commit some files. I am not seeing the eraser tool, and the undo/redo don't seem to work...
It would be awesome if we could work together on polishing and integrating these changes.

<img width="1171" alt="image" src="https://user-images.githubusercontent.com/552629/79316966-638d7380-7f05-11ea-8670-ee80b59cb517.png">

rdbeach:
Hi. Sure, just let me know what you need me to do. Obviously, time allowing.

Here is a link to a running demo: https://lcfprep.com/wb2/wb.html

Probably some kinks to work out, since I've only been working on it on my
system.  I changed the server a bit also, btw.

The eraser tool works, but there is really no way to make a really good
eraser with svg. It tends to get sluggish quickly and is best limited to
"touch ups". That's why I hid it in the popout tool. So people would get
used to using the removal tool instead.

[...]

PR #38, opened 2020-04-26
Base changes

lovasoa:
First PR to integrate @rdbeach  work

See https://github.com/lovasoa/whitebophir/issues/31

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ npm-debug.log
 # Openode
 .openode
 wbo-backup.zip
+
+# Jetbrains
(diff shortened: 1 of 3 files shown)


lovasoa:
The switch to async was made on purpose to make the code more readable. You reverted it in 
b35e020. Is there a reason in particular not to use async ?

finnboeger:
I had a look through the code again and don't see any particular necessity for the change. I believe we should be able to keep it as async/await.

lovasoa:
Thank you !

PR #39, opened 2020-04-26
Style changes

lovasoa:


lovasoa:
We can discuss that, but I would like to keep the new icons if possible. They give wbo an unified style across platforms.

lovasoa:
I am not really convinced by the new favicon svg : 

https://github.com/lovasoa/whitebophir/blob/master/client-data/favicon.svg

vs

https://github.com/finnboeger/whitebophir/blob/style-changes/client-data/favicon.svg

finnboeger:
I see your point and I prefer svgs over emoji characters as well. However I believe we might want to choose different icons in some places. The overall change in the GUI is to have smaller and rectangular tool buttons.
The new icons didn't look quite as good at a smaller scale:
old: https://i.imgur.com/wQwzxbD.png
new: https://i.imgur.com/QLXRqV8.png

I would replace the straight line, rectangle and hand icons with the previous versions. For the eraser I would go with the skull and crossbones that @rdbeach added later on with the updated eraser tool. I also quite liked being able to see the current size in the menu, however I guess we could implement that via the cursor and keep the current icon there. The search icon would benefit from a more simplistic stile aswell (a single thick border instead of the current slimmer two-stroke one)

Off-Topic: When merging I'd prefer a squashing of the commits, that way we will always have a working state after each commit on the master

client-data/board.html
```diff
@@ -7,6 +7,7 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1" />
 	<link rel="stylesheet" type="text/css" href="board.css" />
 	<script src="/socket.io/socket.io.js"></script>
+	<script src="https://kit.fontawesome.com/0166a4ad2b.js" crossorigin="anonymous"></script>
```
lovasoa, line 10: I would like to avoid requesting external assets...

finnboeger: I see that you removed font-awesome in the past. I'm sure we can do without it.

lovasoa:
> The search icon would benefit from a more simplistic stile aswell (a single thick border instead of the current slimmer two-stroke one)

You are right. I updated the zoom icon in  dd2a574, I hope it's better.

finnboeger:
I increased the overall icon sizes while keeping the same button size and I think they look quite nice now. https://i.imgur.com/Xy5V3I6.png

lovasoa:
> I would replace the straight line, rectangle and hand icons with the previous versions. For the eraser I would go with the skull and crossbones that @rdbeach added later on with the updated eraser tool.

I updated the line and rectangle icons. I think the hand is fine, and the skull doesn't convey the right meaning. As a new user, I would fear that clicking it would clear the whole board.

lovasoa:
> I increased the overall icon sizes while keeping the same button size and I think they look quite nice now. https://i.imgur.com/Xy5V3I6.png

I quite like your new style too ! 

I think there is a problem with color presets, though: 

image

finnboeger:
We might need to change the eraser icon to something different nevertheless. At the moment it is the only eraser so there is no problem but later we will have
- the removal tool (currently eraser)
- the eraser (with the svg masking)
- the "clear the board" tool (which should get the trashcan icon imho)

Though I think that is a discussion for later

lovasoa:
Ok, I merged it as is, but I'm still going to make a few changes. Thank you !

And yes, we'll need to discuss the other tools separately.

PR #40, opened 2020-04-26
Add an ellipse and circle tool

finnboeger:


lovasoa:
The circlee tool is a great idea, but it should not be merged with the rectangle tool. It's very unintuitive that you would have to click on the rectangle to draw a circle.

finnboeger:
I can split it out into a separate tool.
What is your opinion on having an ellipsis tool as a secondary function of the circle tool (or vice versa)?

lovasoa:
A common pattern (in inkscape, gimp, google slides, ...) is to have a tool that defaults to ellipsis, and switches to circle when a key (usually shift) is pressed. Could we do that ?

circle

lovasoa:
Also, circles should be added to the server-side svg rendering : 

https://github.com/lovasoa/whitebophir/blob/master/server/createSVG.js

finnboeger:
I can do that

finnboeger:
Should be ready to merge

lovasoa:
I think there is a problem. When you draw an ellipse, and then press shift, it turns into a circle. I think this should happen only while you are drawing, not once the ellipse has already been drawn. What do you think @finnboeger ? Was it on purpose, or was it a mistake. If so, can you fix it ?

finnboeger:
That is a mistake, it should of course only toggle the shape while it is still being drawn.

lovasoa:
I noticed several small bugs around the toggling between circle and ellipse. I'm working on it now, hopefully I'll get to something mergeable today.

finnboeger:
With the new version I encounter several bugs:
- when toggling the tool to the circle state it reverts back to the ellipse mode as soon as I start drawing
- when pressing shift to change to the circle state it does not change until I move the mouse again and force a draw

The previous behavior where toggling the tool via the tool button changes the primary function (what is being drawn when shift is depressed) was intentional as this is the only way that you can draw a circle on mobile devices.

lovasoa:
@finnboeger Patience !  I'm almost done. I'll let you know when you can review.

lovasoa:
@finnboeger, I think this is ready. If you're OK, I'll merge this.

finnboeger:
Looks great :+1: 

Thank you so much for your time and effort

lovasoa:
Thanks to you, @finnboeger ! I think this project has never been so active 🎊

Issue #8, opened 2018-12-02
Give the eraser a width


lovasoa:
Currently, the eraser is just one pixel wide. It means that you have to be very precise when trying to erase a small shape, like a single point. We could make the eraser larger, so that you can just click less that X pixels away from a shape to erase it.

adrian1001:
For long pencil strokes deleting the entire object is usually a bit too much.
Boils down to what is already mentioned in the eraser.js  `//TODO: add the ability to erase only some points in a line`.

Thoughts:
- Add function to boolean-subtract a shape from the drawing (there must be many libraries, e.g. https://github.com/w8r/GreinerHormann, or https://sourceforge.net/p/jsclipper/wiki/Home%206/)
- Add rectangle-eraser: click to mark corner of rectangle, drag to diagonal corner, upon release clip the drawing against rectangle
- Add a pen-like eraser:
  - option1: At every position, clip drawing against a circle-like polygon around cursor position
  - option2 (optimized): Tool draws a white shape to give user feedback; upon release, the shape is used to clip the drawing

Potential issues:
- performance: clipping a large drawing might require good algorithms (partitioning)
- cutting a line into segments will not look nice for thick lines because the end points have no information about at what angle the line was cut. (Draw line as polygon?)

lovasoa:
Yes, these are good ideas. But all of them require a significant refactoring of the existing code. There is currently no spacial index of the shapes.

PR #41, opened 2020-04-26
Add bulk deletion variant of eraser

finnboeger:
Again code mainly by @rdbeach.
Adds secondary function to eraser where a rectangle can be drawn and everything that is fully inside this rectangle will get deleted.


lovasoa:
What is this 2d.js ? I don't have anything against using an external library, but if we do so, then we should probably switch the frontend to use some kind of compilation system, and use npm for dependencies. And we have to make sure any library we use is compatible with the GPL license of this repository.

Making the frontend to a proper web application, using npm, adds some overhead, but I think it's with it. What do you think ?

finnboeger:
I'm not 100% sure, I believe it is used in the masking eraser. @rdbeach Can you explain it? Otherwise I'd work around it and implement necessary functions myself. I have removed it for now.

finnboeger:
I have adjusted the `erase` function to check all points within a given radius (determined by the pencil size). This fixes #8
In conjunction with a pointer that also uses the pencil size it should work quite well

rdbeach:
D2 is not used for the masking eraser. It's used for the transform tool. The library is from http://www.kevlindev.com/. It had some capabilities to put handles on objects and transform them a bit. I used it as a starting point for making the transform tool. With a bit of reworking, you could probably get rid of it.

lovasoa:
Good that we could get rid of 2d.js. I find the tool a little bit unintuitive as it is. I think we should completely remove the original delete on hover behavior, and the tool should just :
 - on click: delete the clicked element, if any
 - on drag: start a selection

finnboeger:
I quite like the current delete function to quickly delete handwritten text. With the increased width of the eraser it works a lot better than it currently does. It would benefit from a different UI that better explains what each tool does.

client-data/board.html
```diff
@@ -87,9 +87,11 @@
 	<script src="../tools/rect/rect.js"></script>
 	<script src="../tools/text/text.js"></script>
 	<script src="../tools/eraser/eraser.js"></script>
+	<script src="../tools/multi-eraser/multi-eraser.js"></script>
 	<script src="../tools/hand/hand.js"></script>
 	<script src="../tools/zoom/zoom.js"></script>
 	<script src="../js/canvascolor/canvascolor.min.js"></script>
+>>>>>>> lovasoa/whitebophir
```
lovasoa, line 94: Whoops, looks like a conflict that was committed by mistake :smile: 

```suggestion
```

lovasoa:
The svg standard defines a function to list the elements inside a rectangle: [`getIntersectionList`](https://www.w3.org/TR/SVG11/struct.html#__svg__SVGSVGElement__getIntersectionList). Unfortunately, browser support isn't there yet (there is an [11 years old bug report in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=501421)), but we could use it when it's there.

lovasoa:
And if you want to test your changes on large boards, you can download the [ic](https://wbo.ophir.dev/boards/ic) board: http://wbo.ophir.dev/download/ic 

There is also [ic2](https://wbo.ophir.dev/boards/ic2), and [ic3](https://wbo.ophir.dev/boards/ic3) .

finnboeger:
I am currently performing some testing with the ic board. The multi-eraser that deletes everything fully contained in a rectangular area has no issues with large boards, the regular eraser in both the click and drag variety has significant issues, locking up the CPU for several seconds. I will attempt to create a function that can determine whether the cursor is over the path of a svg shape and check if I can get it to be faster than the current solution

Issue #20, opened 2020-03-20
Feature: Adding raster image (from clipboard or upload)


jondo:
optionally reducing its resolution.

lovasoa:
Hello, and thank you for getting in touch  !
  Pull requests are welcome !

lovasoa:
This is now better tracked by #235.

Both issues request adding raster images to a board through paste/upload/drop so users can draw on top of them. #235 has the more current discussion, additional use cases, and a linked WIP implementation (#287), so it is the better canonical issue for this workstream.

Keeping #235 open and closing this older issue as a duplicate.

PR #42, opened 2020-04-26
Added document tool to upload images

finnboeger:


client-data/tools/document/document.js
```diff
@@ -3,33 +3,47 @@
...
                 image.onload = function () {
 
                     var uid = Tools.generateUID("doc"); // doc for document
-                    console.log(image.src.toString().length);
 
-                    var msg = {
+                    // File size as data url, approximately 1/3 larger than as bytestream
+                    //TODO: internationalization
+                    let size = image.src.toString().length;
+                    if (size > 1048576) { //TODO: get correct size from config
```
finnboeger, line 28: @lovasoa what would be the best way to access `config.MAX_DOCUMENT_SIZE` from here? I don't want to expose all values set in the config here.

finnboeger:
This also fixes #20. Currently there is no size/quality reduction, however it could be done using e.g. mozjpeg.

client-data/tools/document/document.js
lovasoa: We could reuse the same mechanism we have for translations to pass the configuration to the clients.

lovasoa:
The size/quality reduction could be done prior to sending on the client side with a canvas, with the server only validating that the document does not exceed the maximum size.

server/sockets.js
```diff
@@ -96,6 +97,25 @@ function socketConnection(socket) {
...
+				let existingDocuments = 0;
+				for (key in boardData.board) {
+					if (boardData.board[key].type === "doc") {
+						existingDocuments += 1;
+					}
+					if (existingDocuments >= config.MAX_DOCUMENTS) {
+						console.warn("Received too many documents");
+						return;
+					}
+				}
```
lovasoa, line 111: The number of attached documents should probably be kept in the `BoardData` object (in `boardData.js`), and not recomputed from scratch on every document upload.

server/sockets.js
```diff
@@ -96,6 +97,25 @@ function socketConnection(socket) {
...
+			});
+
+			if (!message.data.size || message.data.size > config.MAX_DOCUMENT_SIZE) {
```
lovasoa, line 114: `.length`, not `.size`, no ? This would have been a security issue. It may be a good idea to add a test in `integration.js` to test that large documents are indeed refused.

finnboeger: Size is a property of the message but I agree that it should be replaced by `message.data.data.length` to actually combat a malicious message

server/sockets.js
finnboeger: I have been hesitant to change the way the data is stored so far but I see the need for some metadata storage. This would be useful for e.g. the preview and export functionality where we could set `lastmodified` and `lastgenerated` attributes to prevent us from generating the svg anew every time.

How about modifying the structure as follows?
```
{
   docCount: ...
   data: ...
}
```

lovasoa: I'd like to avoid breaking backwards compatibility by changing the file format. We may come to it at some point, but we'll need to take the time to discuss the design in depth before. What I was suggesting above was to keep the document count in memory in the `BoardData` object. The count will be computed on disk load, and then incremented on document addition.

finnboeger: That is a good idea that would also work for my other scenario.

finnboeger:
Option to disable tools added in #66

client-data/tools/document/document.js
```diff
@@ -0,0 +1,103 @@
...
+            throw new Error("Too many documents exist already");
+        }
+    }
+
+    function onstart() {
+        var fileInput = document.createElement("input");
+        fileInput.type = "file";
+        fileInput.accept = "image/*";
+        fileInput.click();
+        fileInput.addEventListener("change", () => {
```
lovasoa, line 19:
```suggestion
        fileInput.addEventListener("change", function () {
```

server/sockets.js
```diff
@@ -96,6 +97,28 @@ function socketConnection(socket) {
...
+			}
+
+			boardData.existingDocuments += 1;
+		} else if (message.data.type === "delete") {
+			boardData = await getBoard(boardName);
+
+			if (boardData.board[message.data.id].type === "doc") {
+				boardData.existingDocuments -= 1;
+			}
+		}
```
lovasoa, line 121: This should probably be handled in the BoardData class

client-data/tools/document/document.js
```diff
@@ -0,0 +1,103 @@
...
+                    var scale = 1;
+
+                    do {
+                        // Todo give feedback of processing effort
+
+                        ctx = document.createElement("canvas").getContext("2d");
+                        ctx.canvas.width = image.width * scale;
+                        ctx.canvas.height = image.height * scale;
+                        ctx.drawImage(image, 0, 0, image.width * scale, image.height * scale);
+                        var dataURL = ctx.canvas.toDataURL("image/webp", 0.7);
```
lovasoa, line 45: We should probably use jpeg by default here... We can make it configurable, though.

https://caniuse.com/#feat=webp

finnboeger: jpeg does not support transparency, which would require us to check what kind of image was uploaded and if it made use of the alpha channel.
If it is a png with transparency we would not be able to compress it other than by shrinking the dimensions.

client-data/tools/document/document.js
```diff
@@ -0,0 +1,103 @@
...
+    function draw(msg) {
+        var aspect = msg.w/msg.h;
+        var img = Tools.createSVGElement("image");
+        img.id=msg.id;
+        img.setAttribute("class", "layer-"+Tools.layer);
+        img.setAttributeNS(xlinkNS, "href", msg.data);
+        img.x.baseVal.value = msg['x'];
+        img.y.baseVal.value = msg['y'];
+        img.setAttribute("width", 400*aspect);
+        img.setAttribute("height", 400);
```
lovasoa, line 89: We should probably have a maximum and a minimum image size, and scale only images that do not fit.

server/sockets.js
lovasoa: Also, we should also verify the provided URL, to check that it's a data URL.

client-data/tools/document/document.js
finnboeger: I would suggest using a webp polyfill, e.g. https://github.com/chase-moskal/webp-hero

server/sockets.js
finnboeger: we would still broadcast the large files to all users in that case

lovasoa: Yes, you are right, I hadn't thought about that. I still think the board-specific code should happen in BoardData, but we should probably set the [`httpBufferSize`](https://socket.io/docs/server-api/#new-Server-port-options) option of socket.io in addition, to avoid having to handle and dispatch large messages.

client-data/tools/document/document.js
lovasoa: Let's maybe start with JPEG, and then add support for webp, with a polyfill, in a second pull request ? webp-hero won't work as-is with data urls...

finnboeger:
I plan on adding general support to move and resize shapes which would include images as well.
However the earliest I expect to have time to continue working on this would be in close to a month.

lovasoa:
Doesn't the existing hand tool work to move images created with this patch?

finnboeger:
I am not up to speed regarding the tools that have been added since my last commit so I can't comment on that right now

TechRova:
> This also fixes #20. Currently there is no size/quality reduction, however it could be done using e.g. mozjpeg.

The demo website dont have image adding feature

stranljip:
I would love to see this feature. Is this issue still in progress or has it become stale?

lovasoa:
I just haven't had the time to work on it yet.

lordneeko:
In addition, the ability to "paste" into the window and have it show up within the canvas would be desirable. Perhaps that is a follow-on PR.

Y0ngg4n:
any updates here?

Keraldi:
Hi, I'd also really like this feature. Can anything be done to help the progress?

gigabyteservice:
this will produce an error with downloading functionality/tool

mrducky4:
> this is producing an error with downloading functionality/tool

@gigabyteservice I saw a problem with this, the downloaded SVG would not open if the board included an uploaded image. I fixed it by adding the attribute `xmlns:xlink="http://www.w3.org/1999/xlink"` to the `<svg id="canvas">` element in board.html.

client-data/tools/document/document.js
```diff
@@ -0,0 +1,103 @@
...
+                                Tools.server_config.MAX_DOCUMENT_SIZE / size
+                        ));
+                    } while (size > Tools.server_config.MAX_DOCUMENT_SIZE);
+
+                    var msg = {
+                        id: uid,
+                        type: "doc",
+                        data: dataURL,
+                        size: size,
+                        w: this.width * scale || 300,
```
mrducky4, line 62: This uses the newly reduced value for `scale`. I think it was meant to use the previous value, which is the value used with `drawImage()`

server/boardData.js
```diff
@@ -231,7 +231,11 @@ BoardData.load = async function loadBoard(name) {
 	try {
 		data = await fs.promises.readFile(boardData.file);
 		boardData.board = JSON.parse(data);
-		for (id in boardData.board) boardData.validate(boardData.board[id]);
+		boardData.existingDocuments = 0;
+		for (id in boardData.board) {
+			boardData.validate(boardData.board[id]);
+			if (boardData.board[id].type === "doc") existingDocuments += 1;
```
mrducky4, line 237: Should this be `boardData.existingDocuments`?

server/configuration.js
```diff
@@ -26,6 +26,12 @@ module.exports = {
     /** Maximum value for any x or y on the board */
     MAX_BOARD_SIZE: parseInt(process.env['WBO_MAX_BOARD_SIZE']) || 65536,
 
+    /** Maximum size of uploaded documents default 1MB */
+    MAX_DOCUMENT_SIZE: parseInt(process.env['WBO_MAX_DOCUMENT_SIZE']) || 1048576,
```
mrducky4, line 30: Large images failed for me, causing a transport error. The socket.io default maxHttpBufferSize is 1e6, so I changed this to 1e6 - 500.

NyaomiDEV:
is there still some interest in merging this PR?

holema:
@lovasoa and @NyaomiDEV , it would be great to have the feature. I`m still missing this feature so much

PR #44, opened 2020-04-26
Add oneTouch option when creating tool. 

finnboeger:
These tools won't be selected, rather their onstart action will be executed once.

Useful for undo/redo/clear/upload/zoom-in/zoom-out


lovasoa:
Merged into #61.

PR #45, opened 2020-04-26
Improve text tool

finnboeger:
Text input field appears under the cursor now.
Tool size gets updated when clicking into existing text field

### DIFF
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -511,7 +511,7 @@ Tools.setSize = (function size() {
 
 	chooser.onchange = chooser.oninput = update;
 	return function (value) {
-		if (value) { chooser.value=value; update(); }
+		if (value !== null && value !== undefined) { chooser.value=value; update(); }
 		return parseInt(chooser.value);
 	};
 })();
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -501,15 +501,22 @@ Tools.getColor = (function color() {
 
 Tools.colorPresets.forEach(Tools.HTML.addColorButton.bind(Tools.HTML));
 
-Tools.getSize = (function size() {
+Tools.setSize = (function size() {
 	var chooser = document.getElementById("chooseSize");
+
 	function update() {
 		chooser.value = Math.max(1, Math.min(50, chooser.value | 0));
 	}
 	update();
+
 	chooser.onchange = chooser.oninput = update;
-	return function () { return chooser.value; };
+	return function (value) {
+		if (value) { chooser.value=value; update(); }
+		return parseInt(chooser.value);
+	};
 })();
+
+Tools.getSize = (function() {return Tools.setSize()});
 
 Tools.getOpacity = (function opacity() {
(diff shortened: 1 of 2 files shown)


client-data/js/board.js
```diff
@@ -501,16 +501,23 @@ Tools.getColor = (function color() {
...
+
 	function update() {
 		chooser.value = Math.max(1, Math.min(50, chooser.value | 0));
 	}
 	update();
+
 	chooser.onchange = chooser.oninput = update;
-	return function () { return chooser.value; };
+	return function (value) {
+		if (value) { chooser.value=value; update(); }
```
lovasoa, line 514:
```suggestion
		if (value!=null) { chooser.value=value; update(); }
```

client-data/tools/text/text.js
```diff
@@ -133,6 +167,13 @@
 				if (textField === null) {
 					console.error("Text: Hmmm... I received text that belongs to an unknown text field");
 					return false;
+				} else {
+					/*if(Tools.useLayers){
+						if(textField.getAttribute("class") !== "layer"+Tools.layer){
+							textField.setAttribute("class","layer-"+Tools.layer);
+							Tools.group.appendChild(textField);
+						}
+					}*/
```
lovasoa, line 176: can we remove this ?

client-data/tools/text/text.js
```diff
@@ -151,6 +192,9 @@
 		elem.id = fieldData.id;
 		elem.setAttribute("x", fieldData.x);
 		elem.setAttribute("y", fieldData.y);
+		/*if(Tools.useLayers) {
+			elem.setAttribute("class", "layer-" + Tools.layer);
+		}*/
```
lovasoa, line 197: and this

client-data/tools/text/text.js
finnboeger: I will reintroduce it when I implement the masking eraser. Having it already present will make that step faster, although I can of course just commit it again

client-data/tools/text/text.js
finnboeger: same here

client-data/js/board.js
finnboeger: Why would you only check `null`? We should also cover `undefined`. And since `0` is not a valid size anyway this check should work.

lovasoa: `!=null` covers undefined. This makes it explicit that we are checking for the case where the function was called without arguments. And this changes the behavior of `setSize(0)` to use the minimum size.

client-data/tools/text/text.js
lovasoa: Yes, we can re-add that later if we introduce layers, but currently, we shouldn't have commented-out code.

client-data/js/board.js
finnboeger: I prefer not having any type coercions (except for checking if a value is true-ish which tends to be intentional). If we check for `null` and `undefined` we should make it explicit.

PR #46, opened 2020-04-26
Add a cursor

finnboeger:
Everybody gets a cursor with the color and size they are currently using.
Updates are throttled to 60/second to avoid congestion.
Idle cursors are removed after 3 Minutes (as soon as a different cursor moves).


lovasoa:
Unfortunately, we cannot use es6 syntax if we don't add a javascript compiler. We should either switch to a proper build system, or drop the `let` and `const`.

server/sockets.js
```diff
@@ -2,8 +2,8 @@ var iolib = require('socket.io')
 	, log = require("./log.js").log
 	, BoardData = require("./boardData.js").BoardData;
 
-var MAX_EMIT_COUNT = 64; // Maximum number of draw operations before getting banned
-var MAX_EMIT_COUNT_PERIOD = 5000; // Duration (in ms) after which the emit count is reset
+var MAX_EMIT_COUNT = 200; // Maximum number of draw operations before getting banned
+var MAX_EMIT_COUNT_PERIOD = 1000; // Duration (in ms) after which the emit count is reset
```
lovasoa, line 6: 200 messages per second per user is a lot to handle for a server ! I think a dozen cursor updates per second are okay...

finnboeger: We should let this be configured in `configuration.js`. The cursor alone can send up to 60 messages per second with the currently used tool adding the same (assuming we throttle them as well). 
I have not been able to perform stress tests with a lot of connected devices so I don't know how large the performance impact is.
@rdbeach you currently run your server with a limit of 2000 messages per second per user and with no throttling on the cursor updates. How much impact is it actually?

finnboeger:
I have removed `let` and `const` from the client side code and am pulling the emit count data from the configuration. I would still like to see some performance numbers with multiple users to be able to recommend some default settings depending on the use case. E.g. a private server used for tutoring can afford to have a snappier experience than a large public server

lovasoa:
I started testing it locally. I found issues with the eraser. The cursor starts blinking when the eraser is used.
blink

lovasoa:
Also: the cursor can be erased with the eraser tool. It shouldn't. It should probably not be part of the svg element that contains the board at all.

lovasoa:
Oh, and when hovering over the menu, then giving the focus to another window, then given the focus back to the window, the cursor disappears:

nocursor

lovasoa:
And we could make the cursor "snappier" without increasing the update frequency by setting a CSS transition on its position. Would that work ?

finnboeger:
That would only be for the local update not for the network updates or am I missing something?

I will have a look at the other issues with the eraser

lovasoa:
The CSS transition would apply to any update to the position of the element. It wouldn't require any javascript.

finnboeger:
I fixed the eraser issue but I wasn't able to reproduce the disappearing cursor in either Firefox or Google Chrome. 

As for the css transitions, could you do that? I'm not sure whether I completely understand your idea.

client-data/tools/eraser/eraser.js
```diff
@@ -48,7 +48,7 @@
 			var touch = evt.touches[0];
 			target = document.elementFromPoint(touch.clientX, touch.clientY);
 		}
-		if (erasing && target !== Tools.svg) {
+		if (erasing && target !== Tools.svg && target.className.baseVal.indexOf("opcursor") === -1) {
```
lovasoa, line 51: I don't think the eraser tool should have to be aware of the pointer...
We could have a `<g>` element inside of which all the board elements are, and let the eraser check for that instead.

lovasoa:
Okay, I'll try to work on this during this week.

client-data/tools/eraser/eraser.js
finnboeger: There is a fast way to do this using `Node.contains()` however IE does not support this for svg elements. Do you want to support IE or can I ignore it?

server/sockets.js
rdbeach: I have the luxury of just having a few people on my board at a time, so I admit I haven't as yet put a lot of effort into efficiency/stress testing.

That being said, I just updated the pointer with your suggestion of throttling. Currently 20 updates per second for the pointer. Also, I changed it so that when you are drawing there are no separate pointer messages. 1ad204b4e28bade65c9296a904f316a25b6c396c

@lovasoa how does the board hold up with several users/several boards running simultaneously?

Also, one issue I had with the original board was that the pencil was throttled too much to my liking, so that the reproduction on the other persons board was noticeably different in some cases. 

`
if (curLineId !== "" && performance.now() - lastTime > 70) {\n\r
	Tools.drawAndSend(new PointMessage(x, y));
	lastTime = performance.now();
}
`
So I reduced the throttle time, but then I had to raise the MAX_EMIT_COUNT to prevent getting banned.

One more issue with the code above is that sometimes the final (x,y) position before the mouseup  event is skipped or lost, and since many people do a quick stroke before they lift their pencil when writing, some of the character and clarity of the letters is lost. To deal with this, I made sure that the final mouse position was sent, overriding the throttle.  Still, at 70ms, handwriting will not be reproduced faithfully, reducing this to 20-30 seems to do the trick, of course, at the expense of bandwidth.

finnboeger: I have implemented your change to the pencil tool in #47.
Not sending the cursor updates when using a tool is a great idea, I will have a look at your commit and add it to this PR.

finnboeger:
I have changed the cursor positioning from updating x/y coordinates to updating the transform translate property which should improve performance

lovasoa:
Also, I don't know if you noticed, but the cursor now appears *behind* elements that are on the board.

server/sockets.js
lovasoa: Ideally, MAX_EMIT_COUNT would be passed to the client, which could use it to compute its own optimal throttling. For the pen, these are very interesting improvements !

finnboeger:
I did indeed miss that, should be fixed now.

lovasoa:
Screenshot 2020-04-28 at 17 08 57


I think there is a bug in the management of cursor elements. All cursors have the same ID, and a new cursor element is created for every cursor message received from the server, even if they come from the same user.

finnboeger:
Fixed the cursor handling bug (outsourced cursor creation into a function but then ignored the parameters...)

Passing parts of the server config through to the client and computing optimal throttling from it now.

server/sockets.js
finnboeger: Batch processing is definitely preferable to dropping them, especially since the user would never notice that others can't see them.

I'm wondering whether that is still in the scope of this PR or if we should split that into another one and finish the cursor without the more intelligent throttling for now?

lovasoa: Yes, we should probably do that in a separate PR. This should not be too hard, since there is already a batching mechanism, with `_children`

lovasoa:
👍  for the client config importation. This will be useful in other contexts too.

client-data/tools/eraser/eraser.js
```diff
@@ -39,6 +39,22 @@
 		"type": "delete",
 		"id": ""
 	};
+
+	function inDrawingArea(elem) {
+		if (window.navigator.userAgent.indexOf("MSIE") === -1) {
```
lovasoa, line 44:
```suggestion
		if (Tools.drawingArea.contains) {
```

we don't need UA sniffing

lovasoa:
Ok, I've started working on this, and I hope I'll have something ready for merging by the weekend. I would like to integrate the cursor as a tool instead of duplicating all the position calculation, message sending, message dispatch logic, and so on. It will force us to implement tools without an icon in the menu, and multiple simultaneous active tools, but I still think it will create clutter for the long term.

finnboeger:
Do we really want to add the cursor as a tool? I agree with isolating it but given that it in no way behaves like a tool I think it should instead register its handler directly and spare us the headache of dealing with non-tool "tools".

lovasoa:
Well, we don't want to dupicate any of the event handling code and event position calculation code. We shouldn't have to add any code to handle non-tool tools in board.js, I'll just call the functions I need from cursor.js.

lovasoa:
@finnboeger I think this is ready. Can you have a look ? If you think this is ok, I'll merge.

finnboeger:
Currently when using a tool neither the local cursor nor the remote cursors get updated. I think we should instead still show the local cursor (e.g. for using with the eraser to know the radius of it) but not send it unless the current tool does not constantly transmit a position (again, like with the eraser).
I have committed the necessary changes

client-data/tools/cursor/cursor.js
```diff
@@ -0,0 +1,99 @@
...
+        var message = {
+            type: "update",
+            x: x,
+            y: y,
+            color: Tools.getColor(),
+            size: Tools.getSize(),
+        };
+
+        if (cur_time - lastCursorUpdate > MAX_CURSOR_UPDATES_INTERVAL_MS &&
+            (sending || ["Pencil", "Straight line", "Rectangle"].indexOf(Tools.curTool.name) === -1)) {
```
lovasoa, line 65: This creates a tight coupling from the cursor to other tools' names, which I'd rather avoid. We don't want to have to touch this file when adding new tools or renaming a tool. If you really want this feature, than the tool itself should declare whether it wants the marker to appear, and the check here should look like `Tools.curTool.showMarker`

lovasoa:
Thank you @finnboeger for your work! It's great to see this finally merged :)

finnboeger:
Thank you for all the work you did refactoring the code, it looks a lot better now!

PR #47, opened 2020-04-27
Add a whiteout tool as the secondary function of the pencil tool

finnboeger:
This allows easy touch ups without using the masking eraser. This works best on a white canvas - when using the grid the whiteout spots can be seen.
After #45 has been merged this can be modified to use a different size for the whiteout tool (e.g. twice the current pen size).
After #46 has been merged this should display a marker of the proper size.


lovasoa:
Yes, the eraser should definitely not erase the grid... Also, this should probably not be merged with the pencil...

finnboeger:
This is not an eraser but rather a whiteout tool (like [Tipp-Ex](https://en.wikipedia.org/wiki/Tipp-Ex)) which does overwrite everything below it.

server/sockets.js
```diff
@@ -2,8 +2,8 @@ var iolib = require('socket.io')
 	, log = require("./log.js").log
 	, BoardData = require("./boardData.js").BoardData;
 
-var MAX_EMIT_COUNT = 64; // Maximum number of draw operations before getting banned
-var MAX_EMIT_COUNT_PERIOD = 5000; // Duration (in ms) after which the emit count is reset
+var MAX_EMIT_COUNT = 200; // Maximum number of draw operations before getting banned
+var MAX_EMIT_COUNT_PERIOD = 1000; // Duration (in ms) after which the emit count is reset
```
lovasoa, line 6: I think this is not related to this PR..

lovasoa:
OK for the merging, then...

server/sockets.js
finnboeger: This PR increases the polling rate of the pencil tool leading to a much smoother experience but also an increased message count. I just did some testing and found that I can't exceed 200 messages over 5 seconds unless I actively try. Even attempting to fill out a larger area by rapidly whacking the pencil left and right tends to not exceed that number.

client-data/tools/pencil/pencil.js
```diff
@@ -48,19 +76,19 @@
 		Tools.drawAndSend({
 			'type': 'line',
 			'id': curLineId,
-			'color': Tools.getColor(),
+			'color': (curPen === "pencil" ? Tools.getColor() : "white"),
```
lovasoa, line 79:
```suggestion
			'color': (curPen === "pencil" ? Tools.getColor() : "#ffffff"),
```

for consistency

client-data/tools/pencil/pencil.js
```diff
@@ -38,6 +43,29 @@
...
+				cursor.remove();
+			}
+		}
+	}
+
+	function onStart(){
+		if(curPen === "whiteout"){
+			Tools.showMarker = true;
+		}
+	}
```
lovasoa, line 61: I don't think the cursor should be accessed from other tools. You can use the `showMarker` attribute of individual tools that you introduced in the cursor PR. You can use something like `pencilTool.showMarker = False`

finnboeger: I have decided that since we were displaying the cursor anyway (`alwaysShowMarker` was set to `true`) we don't need any of the code that was checking if it should hide the cursor

lovasoa:
@finnboeger : I made a few changes. If you're okay with them, I'll merge.

lovasoa:
OK, I'll merge, let me know if you have any comment.

finnboeger:
Looks great :+1: 

Sorry for the late response, unfortunately I didn't have time in the last week.

PR #49, opened 2020-04-27
Fix various bugs with the menu

### LINKED ISSUE #48: UI alignment issues is Safari
image

finnboeger:
- Fix the color picker UI issues (Fixes #48)
- Fixes the fact that currently when a tooltip is opened you cannot click into the area below or above it to close it again
- Disables the automatic closing of the tooltips
- Changes from `:hover` to `:focus` on mobile and blurs focus on tool input (basically on click into canvas)
- Close settings sliders automaticallyon desktop after making an adjustment.

### DIFF
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -426,16 +426,10 @@ Tools.toolHooks = [
 		}
 
 		function wrapUnsetHover(f, toolName) {
-			// test for mobile device
-			// https://stackoverflow.com/a/24600597
-			if (/Mobi|Android/i.test(navigator.userAgent)) {
-				return (function unsetHover(evt) {
-					document.activeElement.blur();
-					return f(evt);
-				});
-			} else {
-				return f;
-			}
+			return (function unsetHover(evt) {
+				document.activeElement && document.activeElement.blur();
+				return f(evt);
+			});
 		}
 
 		if (listeners.press) {
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -425,9 +425,22 @@ Tools.toolHooks = [
 			});
 		}
 
+		function wrapUnsetHover(f, toolName) {
+			// test for mobile device
+			// https://stackoverflow.com/a/24600597
+			if (/Mobi|Android/i.test(navigator.userAgent)) {
+				return (function unsetHover(evt) {
+					document.activeElement.blur();
+					return f(evt);
+				});
+			} else {
+				return f;
+			}
+		}
+
 		if (listeners.press) {
-			compiled["mousedown"] = compile(listeners.press);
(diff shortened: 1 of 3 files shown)


lovasoa:
> Changes from :hover to :focus on mobile and blurs focus on tool input (basically on click into canvas)

I'm not sure I understand this... What does that mean ? Why trying to detect mobile browsers in particular ?

client-data/js/board.js
```diff
@@ -425,9 +425,22 @@ Tools.toolHooks = [
 			});
 		}
 
+		function wrapUnsetHover(f, toolName) {
+			// test for mobile device
+			// https://stackoverflow.com/a/24600597
+			if (/Mobi|Android/i.test(navigator.userAgent)) {
+				return (function unsetHover(evt) {
+					document.activeElement.blur();
```
lovasoa, line 433: activeElement can be null, in which case this will throw an error.

client-data/js/board.js
```diff
@@ -425,9 +425,22 @@ Tools.toolHooks = [
 			});
 		}
 
+		function wrapUnsetHover(f, toolName) {
+			// test for mobile device
+			// https://stackoverflow.com/a/24600597
+			if (/Mobi|Android/i.test(navigator.userAgent)) {
```
lovasoa, line 431: If this is a screen size problem, then we should probably detect the screen size, not whether this is a mobile browser or not. User-Agent sniffing is generally not a good idea...

finnboeger:
When you click on a tool in a mobile device the internal cursor of the browser hovers above it. Since we prevent the click events on the svg from bubbling up by calling `evt.preventDefault` in most of the tools there is never a click somewhere else and thus the browser believes that we are still hovering over the menu and the tooltip never collapses.
Removing the `evt.prepareDefault` statements breaks more stuff so I went looking for an alternative. By setting a `tab-index` we can allow the menu buttons to be focused. Now when a user clicks on them they become focused and if the browser does not properly supports hovering, as determined by the media query, the css falls back on using the focus state to display the tooltip. We can then blur the active element if any tool gets used which happens whenever the user clicks somewhere that is not the menu. If he clicks on a different menu button that one get focused which has the same effect.
We don't really need to check the user agent and can run the code on desktop as well but it would be redundant there as we don't use the focus state.

As for `activeElement` being `null`: it defaults to `body` in chrome but I haven't tested it in other browsers, I will add a guard there since it doesn't hurt.

lovasoa:
OK, I'm merging this because it's already a great improvement over what we currently have. However, there are still issues with the size selector not closing properly on mobile.

I tested both firefox mobile and chrome mobile, both have the CSS4 media query  `hover: hover`.
https://mediaqueriestest.com/

PR #50, opened 2020-04-27
Improve scroll zoom handling

finnboeger:
This fixes a bug where if you scroll-zoom out, no matter how fare zoomed in you were, the scale gets set to 0.1
Now all browsers should zoom with the same step size


lovasoa:
I think I had seen this bug before, without managing to pinpoint what was causing it. Thank you !

PR #52, opened 2020-04-28
catch mobile devices that misreport their hover capabilities

### LINKED ISSUE #51: On mobile, the size selector does not close when it is blurred
The size selector stays open when the user starts drawing on mobile browsers (tested in chrome and firefox mobile).

See #49

finnboeger:
Fixes #51

### DIFF
--- a/client-data/board.css
+++ b/client-data/board.css
@@ -113,7 +113,7 @@ html, body, svg {
 	max-width: 100%;
 }
 
-@media (hover: none) {
+@media (any-hover: none) or (any-pointer: coarse) {
 	#menu .tool:hover {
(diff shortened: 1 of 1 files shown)


client-data/board.css
```diff
@@ -113,7 +113,7 @@ html, body, svg {
 	max-width: 100%;
 }
 
-@media (hover: none) {
+@media (any-hover: none) or (any-pointer: coarse) {
```
lovasoa, line 116:
```suggestion
@media (hover: none), (pointer: coarse) {
```

I don't think there is a `or` in CSS media queries. And using the data reported for the main input device is probably more appropriate.

PR #54, opened 2020-04-29
Add option to change pen size with the scroll wheel

finnboeger:
When scrolling while holding the `alt` key the pen size will be increment or decremented.
When also holding the `shift` key, the change in size will be smaller to allow for finetuning.


client-data/tools/zoom/zoom.js
```diff
@@ -82,14 +82,22 @@
...
             var scale = Tools.getScale();
             var x = evt.pageX / scale;
             var y = evt.pageY / scale;
             setOrigin(x, y, evt, false);
             animate((1 - ((evt.deltaY > 0) - (evt.deltaY < 0)) * 0.25) * Tools.getScale());
+        } else if (evt.altKey) {
+            // make finer changes if shift is being held
+            var change = evt.shiftKey ? 1 : 5;
+            // change tool size
+            Tools.setSize(Tools.getSize() - ((evt.deltaY > 0) - (evt.deltaY < 0)) * change);
```
lovasoa, line 95: Couldn't we make it directly proportional to the deltaY ?

finnboeger: That would be different in each browser (same issue that the scroll to zoom function had initially). Firefox reports a delta of 3 whereas chrome reports 53 (at least on my device).

lovasoa:
I updated the code to give a visual feedback when changing the cursor size. Let me know what you think.

finnboeger:
lgtm :+1:

PR #55, opened 2020-04-30
change url to hide app behind a relative path

lmazet:
This changes are needed to hide the nodejs service behind a web server with a relative path.

PR #57, opened 2020-04-30
Allow running the app behind a reverse proxy

### LINKED ISSUE #32: links with relative path
Hi,

I try to deploy wbo with docker, behind an apache webserver with proxypass because I want to use it as a subsection of a website : https://www.toto.fr/wbo

But some resources are in absolute path, when I try to access https://www.toto.fr/wbo/boards/test, /background.png, /socket.io/socket.io.js, /js/canvascolor/canvascolor.min.js gets 404 or 500 error code.

Or maybe I made a mistake in my apache conf ?

```
<Location /wbo>
        ProxyPass "http://wbo.host.fr:5001/"
        ProxyPassReverse "http://wbo.host.fr:5001/"
</Location>
```

finnboeger:
Based on #55. 
Fixes #32.

Adds a new environment variable (`WBO_URL_PREFIX_PATH`) to configure at which path the code will be run, defaults to domain root.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -27,7 +27,7 @@ log("server started", { port: config.PORT });
 var CSP = "default-src 'self'; style-src 'self' 'unsafe-inline'; connect-src 'self' ws: wss:";
 
 var fileserver = function(request, response, errorHandleFunction) {
-	request.url = request.url.replace(config.URL_PREFIX_PATH, "");
+	if (config.URL_PREFIX_PATH) request.url = request.url.replace(config.URL_PREFIX_PATH + "/", "");
 	return serveStatic(config.WEBROOT, {
 		maxAge: 2 * 3600 * 1000,
 		setHeaders: function (res) {
@@ -72,7 +72,8 @@ const indexTemplate = new templating.Template(path.join(config.WEBROOT, 'index.h
 
 function handleRequest(request, response) {
 	var parsedUrl = url.parse(request.url, true);
-	var parts = parsedUrl.pathname.replace(config.URL_PREFIX_PATH, "").split('/');
+	var path = config.URL_PREFIX_PATH ? parsedUrl.pathname.replace(config.URL_PREFIX_PATH + "/", "") : parsedUrl.pathname;
+	var parts = path.split('/');
 	if (parts[0] === '') parts.shift();
 
 	if (parts[0] === "boards") {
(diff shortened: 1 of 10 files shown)


server/server.js
```diff
@@ -26,13 +26,16 @@ log("server started", { port: config.PORT });
...
 
-var fileserver = serveStatic(config.WEBROOT, {
-	maxAge: 2 * 3600 * 1000,
-	setHeaders: function (res) {
-		res.setHeader("X-UA-Compatible", "IE=Edge");
-		res.setHeader("Content-Security-Policy", CSP);
-	}
-});
+var fileserver = function(request, response, errorHandleFunction) {
+	request.url = request.url.replace(config.URL_PREFIX_PATH, "");
```
lovasoa, line 30: We should probably avoid to mutate the request object here

lovasoa: And I'm not sure I understand why this would have to be changed... The front-end reverse proxy should already be rewriting paths, so you shouldn't be receiving anything different here than in the normal case. Or are you trying to get the server to serve the app directly on a different path, withouth a front-end reverse proxy ?

finnboeger: This is to run it directly on a different path or rather behind a reverse proxy without rewriting the paths.

Where would you change the request url instead?

lovasoa: What exactly is the interest of running it directly under a different path without a reverse proxy ? The interest of running it under a path is to be able to run other servers on other paths, isn't it ?

finnboeger: Yes, the only difference it would make is not having to create rewrite rules in the proxy.
I initially assumed that we would need to specify the prefix anyway in the server configuration however I've come to think that this is not the case and it would work with the reverse proxy rewriting the urls as well.

lovasoa: OK, so let's just simplify the code, and create a [wiki page](https://github.com/lovasoa/whitebophir/wiki) explaining how to set up the reverse proxy.

finnboeger: I have attempted to set up nginx to rewrite the url. While that part was successful and I grabbed the prefix out of `window.location.pathname` to get the correct url for socket.io to connect to, I was unable to get a correct handshake.

finnboeger: I have also just noticed that we still need the prefix path to be set to be able to deliver correct absolute URLs for the open graph metadata

lovasoa: So, can you remove the special handling in the file server and socket.io server ?

finnboeger: I have removed the need for setting the prefix and got it running under nginx, I will add my configuration and the example given by @lmazet to the wiki

PR #58, opened 2020-04-30
Add a svg export with proper rounding

### LINKED ISSUE #37: Feature: SVG download for embedding
For embedding whiteboards (e.g. in wikis) it would be nice to have the possibility to download a full SVG, e.g. with a link like /embed/board-name.

This can already be achieved with /preview/board-name, but there all paths are rendered as lines instead of curves - I guess that's intended for a "preview"?

To make it more scalable a simple caching might be wise. Trivial approach would be that createSVG.renderBoard() checks if there is an .svg in server-data/ and otherwise regenerates and saves one, and whenever board-name.json is rewritten any existing .svg is deleted.

finnboeger:
Add server side rounding to the pencil tool for exporting the svg exactly as it is displayed

Fixes #37.

<sub>
</sub>


server/createSVG.js
```diff
@@ -58,29 +64,103 @@ var Tools = {
...
+						]
+					};
+					pts.push(npoint);
+				}
+				for (let i = 1; i < pts.length; i++) {
+					pathstring += " " + pts[i].type + " " + pts[i].values.join(" ");
+				}
+		}
+
+		return renderPath(el, pathstring);
```
lovasoa, line 142: Can't we refactor all this code in order to use the same code on the client and the server ?

lovasoa:
It looks like this broke the tests

finnboeger:
I changed the handling of paths with only two points to draw a line instead of a bezier curve which broke the test.
Do we want to keep drawing them as curves or change the test and introduce a new one for paths with more points?

server/createSVG.js
finnboeger: How would you go about accessing the function? We can't use `require` on the client side but we also don't want to pollute our namespace on the client side either.

client-data/tools/pencil/pencil.js
```diff
@@ -118,9 +118,11 @@
 		switch (nbr) {
 			case 0: //The first point in the line
 				//If there is no point, we have to start the line with a moveTo statement
-				npoint = { type: "M", values: [x, y] };
+				pts.push({ type: "M", values: [x, y] });
+				//Temporary first point so that clicks are shown and can be erased
+				npoint = { type: "L", values: [x, y] };
 				break;
-			case 1: //There is only one point.
+			case 2: //There are two points, however one of them might be temporary. Since it is a line of 0 length we can ignore it
```
lovasoa, line 125: I don't understand this comment. Why *might* the point be temporary ?

lovasoa:
Yes, I think we should have tests for straight line and bezier curve, and check the "preview" in each.

client-data/tools/pencil/pencil.js
finnboeger: There should never be a case where there is only one point since if there are no points, two will be created. I guess there could be old boards that have only a single child that we might need to deal with.
I initially planned for the first point to be deleted later but found that to be impossible (we create the line and have two points -> we add another point and want to delete the previous point since it is no longer needed -> now we still have 3 points and keep deleting the previous one. Checking the type of the second point did not work as chrome set it to be a line every time).

lovasoa: OK, thanks for the clear explanation !
If we are sure there can never be 1 point, then we should throw an error in that case, otherwise, we should probably just handle it as before.

finnboeger: I will update it to ensure correct rendering of older boards

finnboeger:
The only thing left to do here is reusing the rendering function on the server side. Any tipps as to how are greatly appreciated.

A future addition would be adding some place to store metadata (when was the board last updated & when was the last preview/export generated), storing the generated svgs and only regenerating them if something has changed or possible only after some time has passed as well in order to prevent overloading the server.

lovasoa:
The quick solution is to just create a file in client-data, include it from board.js and require it from the server. The nice solution is to switch to a proper build system. What do you think ?

finnboeger:
I would say quick and dirty for now and then clean it up properly when we got the current PRs merged?

lovasoa:
> I would say quick and dirty for now and then clean it up properly when we got the current PRs merged?

OK !

tests/integration.js
```diff
@@ -24,26 +24,63 @@ function testPencil(browser) {
...
                 done();
             }, 100);
         })
-        .assert.visible("path[d='M 100 200 C 100 200 300 400 300 400'][stroke='#123456']")
+        .assert.visible("path[d='M 100 200 L 100 200 C 100 200 300 400 300 400'][stroke='#123456']")
+        .assert.visible("path[d='M 200 100 L 200 100 C 200 100 252.85954792089683 200 300 200 C 347.14045207910317 200 400 100 400 100'][stroke='#abcdef']")
         .refresh()
-        .assert.visible("path[d='M 100 200 C 100 200 300 400 300 400'][stroke='#123456']")
+        .assert.visible("path[d='M 100 200 L 100 200 C 100 200 300 400 300 400'][stroke='#123456']")
+        .assert.visible("path[d='M 200 100 L 200 100 C 200 100 252.85954792089683 200 300 200 C 347.14045207910317 200 400 100 400 100'][stroke='#abcdef']")
```
lovasoa, line 73: Isn't there a rounding problem here, with `252.85954792089683` ?

finnboeger: I haven't investigated how the code creates those extrapolated points for the bezier curves so I can't comment on that. Instinctively however I would have expected `250` and `350` instead of the current floating point values.

lovasoa:
@finnboeger : I think I'm done. Let me know what you think.

finnboeger:
Looks great!
I found and fixed a small issue with rectangles (since they were being displayed using paths and paths have rounded ends and corners, the rectangles had rounded corners as well)

Two other things:
1. Do we want to display ellipses using the dedicated `ellipse` tag as well and
2. Do we want to cache boards?

Probably should do that in a separate PR though.

lovasoa:
Thanks for the fix ! I'm going to merge it like that, to let us focus on the other PRs. In the long run, we may want to find a way to unify rendering between the front and the back-end, to avoid duplicating CSS and drawing logic for every tool. But this is part of a larger refactoring that we can keep for later. 

I'm not sure exactly what you mean by caching the boards, but we could indeed set a proper ETag header to avoid re-rendering boards that we have rendered already. If you are talking about storing them separately, I'm not sure it's worth it... An performance-sensitive installation in production will probably already be behind a caching reverse proxy, which will understand ETags.

PR #59, opened 2020-05-01
correct favicon on board

lmazet:
A favicon declaration is missing in board.html

### DIFF
--- a/client-data/board.html
+++ b/client-data/board.html
@@ -10,6 +10,7 @@
 	<meta name="description" content="{{translations.tagline}}'" />
 	<meta name="keywords"
 		content="{{translations.collaborative_whiteboard}},online,draw,paint,shared,realtime,wbo,whitebophir" />
+	<link rel="icon" type="image/x-icon" sizes="16x16" href="../favicon.ico">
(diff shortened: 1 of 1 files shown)


client-data/board.html
```diff
@@ -10,6 +10,7 @@
 	<meta name="description" content="{{translations.tagline}}'" />
 	<meta name="keywords"
 		content="{{translations.collaborative_whiteboard}},online,draw,paint,shared,realtime,wbo,whitebophir" />
+	<link rel="icon" type="image/x-icon" sizes="16x16" href="../favicon.ico">
 	<link rel="apple-touch-icon" href="../favicon.svg">
 	<link rel="icon" type="image/x-icon" sizes="16x16" href="../favicon.ico">
```
finnboeger, line 15: it is right here :smile:

lmazet: Sorry. I was working on PR#55 when I discovered that the line was missing.... After updating my repo, I added the line without checking that someone had already inserted it.

finnboeger: Don't worry, thank you for contributing anyway :+1:

PR #61, opened 2020-05-02
Add a grid

finnboeger:
Depends on #44.

<sub>
</sub>


client-data/tools/grid/grid.js
```diff
@@ -0,0 +1,86 @@
...
+            createSVGElement("path", {d: "M 300 0 L 0 0 0 300", fill: "none", stroke: "gray", 'stroke-width': "1"})
+        );
+        // dots
+        var dots = createSVGElement("pattern", {id: "dots", width: "30", height: "30", x: "-10", y: "-10", patternUnits: "userSpaceOnUse"});
+        dots.appendChild(createSVGElement("circle", {fill: "gray", cx: "10", cy: "10", r: "2"}));
+
+        var defs = Tools.svg.getElementById("defs");
+        defs.appendChild(smallGrid);
+        defs.appendChild(grid);
+        defs.appendChild(dots);
```
lovasoa, line 65: Maybe we could define the grid in an external SVG file, to keep the code more readable ?

client-data/tools/grid/grid.js
```diff
@@ -0,0 +1,86 @@
...
+
+    var gridContainer = init();
+
+    Tools.add({ //The new tool
+        "name": "Grid",
+        "shortcut": "g",
+        "listeners": {},
+        "icon": "tools/grid/icon.svg",
+        "oneTouch": true,
+        "onstart": toggleGrid,
```
lovasoa, line 82: What would you think of just letting the onstart function return whether the tool wants to be activated or not ? This would allow us not to have special code to handle `oneTouch` tools, and also let the tools themselves be more flexible about when the want/can be activated.

finnboeger: I have encountered several issues trying to use the onstart function for this but it mainly boils down to this:
If we use onstart without changing anything in `board.js` the UI flickers since the tool will get changed anyway and we can only set a timeout to change it back.
If we instead modify call of the onstart function in the change handler to only change the tool if the onstart function returns a trueish value we will still have called the onquit function of the previous tool.
I don't see an elegant solution to this or anything that is less obstrusive than adding the oneTouch property.

client-data/tools/grid/grid.js
finnboeger: unfortunately IE does not support including external svg with `use` which would be the way to keep the the svg out of the js code. 
However I have just tried testing the site in IE11 and noticed several other problems anyway:
- Promises are not supported (used for the `batchCall` function) for which we could use a polyfill
- multiple svgs for the icons are not rendering correctly (probably need to change the viewBox) which is fixable
- `window.scrollX` and `window.scrollY` don't exist, instead we can use `document.documentElement.scrollLeft` and `document.documentElement.scrollTop`
- only the hand tool is usable at all and even then only after I applied the above fixes.

I just went through the history to see when these were introduced and noticed that the `scrollX` one is from 2013 (5bdf547) and the `Promise` is from over 7 months ago (e03852). I think we should just give up on IE given that it isn't supported at the moment anyhow.

lovasoa: Thank you very much for taking the time to test with Internet Explorer and listing issues ! The missing features of Internet Explorer that you list don't seem to be too fundamental to the way WBO works... Do you think we could get a better browser compatibility without having to write too much specialized code just by integrating something like [polyfill-library](https://www.npmjs.com/package/polyfill-library) ? Would you be interested in working on that ? I don't have a windows computer, but I can test things in a VM.

finnboeger: I'm investigating the necessary changes on the [bug/ie-fixes](https://github.com/finnboeger/whitebophir/tree/bug/ie-fixes) branch. I have a Windows VM set up as well for that purpose.

finnboeger: It looks like it is going to take some deep digging to get it working in IE.
The primary issues are as the following:
- the `removeEventListener` call when switching tools is not removing the listeners which leads to malformed messages. If a user selects any tool and then subsequently any different tool both listeners will be active at the same time leading to the a jumbled mess where one tool will attempt to draw stuff with the message for the other tool.
- IE constantly emits `mouseleave` events for some reason making it nigh impossible to draw anything. This can at least be easily fixed using UA sniffing.

finnboeger: Just had another idea and noticed that the issue has nothing to do with the closures but rather Internet Explorer seems to decide on its own whether the `useCapture` option should be set to true. 
By also attempting to remove the listener in capturing mode if the browser is IE we can fix this

PR #63, opened 2020-05-03
Get WBO working in IE

finnboeger:
Tested in IE11.
Not extensively tested in proper Browsers yet.

<sub>
</sub>


client-data/board.html
```diff
@@ -20,6 +20,7 @@
 	{{#languages}}
 	<link rel="alternate" hreflang="{{.}}" href="{{../boardUriComponent}}?lang={{.}}" />
 	{{/languages}}
+	<script src="https://cdn.jsdelivr.net/npm/promise-polyfill@8/dist/polyfill.min.js"></script>
```
lovasoa, line 23: Currently, WBO doesn't have any dependency to external services. This is good for privacy, and ease of deployment behind a firewall. Maybe we can use something like [polyfill-library](https://www.npmjs.com/package/polyfill-library) on the server instead ? This would also have the advantage of providing many other polyfills, to let us use more modern js features in the code, and ensure compatibility with many old browsers.

client-data/js/board.js
```diff
@@ -49,6 +49,8 @@ Tools.showMarker = true;
 Tools.showOtherCursors = true;
 Tools.showMyCursor = true;
 
+Tools.isIE = /MSIE|Trident/.test(window.navigator.userAgent);
```
lovasoa, line 52: Doesn't this trigger for the latest chrome-based IE ?

client-data/js/board.js
```diff
@@ -593,7 +597,20 @@ if (!Math.hypot) {
...
+		Element.prototype.msMatchesSelector ||
+		Element.prototype.oMatchesSelector ||
+		Element.prototype.webkitMatchesSelector ||
+		function(s) {
+			var matches = (this.document || this.ownerDocument).querySelectorAll(s),
+				i = matches.length;
+			while (--i >= 0 && matches.item(i) !== this) {}
+			return i > -1;
+		};
+}
```
lovasoa, line 613: Maybe this could be handled by the polyfill library too ? (see above)

client-data/js/canvascolor/canvascolor.min.js
```diff
@@ -1,2 +1,2 @@
...
...(g,m);(function(){for(var b=0;b<g;b++)for(var a=0;a<m;a++){var c=4*(b+a*g),d=k([b,a]);s.data[c]=d[0];s.data[c+1]=d[1];s.data[c+2]=d[2];s.data[c+3]=255}p.putImageData(s,0,0)})();h.addEventListener("mousemove", function(a){a=l(a);q.style.backgroundColor=c(k(a))},!0);h.addEventListener("click",function(a){a=l(a);a=c(k(a));f(a);d(a)},!0);h.addEventListener("mouseleave",function(){q.style.backgroundColor=t},!0)}(function(){var a=document.createElement("style");a.innerHTML=".canvascolor-container{background-color:black;border-radius:5px;overflow:hidden;width:179px;padding:2px;display:none;}.canvascolor-container canvas{cursor:crosshair;}.canvascolor-history{overflow:auto;}.canvascolor-history > div{margin:2px;display:inline-block;}"; document.head.appendChild(a)})();var r;(function(){r=document.createElement("div");r.className="canvascolor-container";var a=document.createElement("canvas"),c=document.createElement("div");c.className="canvascolor-history";r.appendChild(a);r.appendChild(c)})();for(var w=document.querySelectorAll("input.canvascolor, input[type=color]"),p=0;p<w.length;p++){var u=w.item(p);"color"===u.type&&-1===u.className.split(" ").indexOf("canvascolor")||v(u)}return v})();
```
lovasoa, line 2: 😬 Haha, I didn't even remember we had that. Maybe we can include the uncompressed code instead ? It is written by me, and not used anywhere but in WBO, I think...

client-data/tools/cursor/cursor.js
```diff
@@ -106,7 +106,8 @@
     function draw(message) {
         var cursor = getCursor("cursor-" + (message.socket || 'me'));
         cursor.style.transform = "translate(" + message.x + "px, " + message.y + "px)";
+        if (Tools.isIE) cursor.setAttributeNS(null, "transform", "translate(" + message.x + " " + message.y + ")");
```
lovasoa, line 109: Is there a reason we don't use the circle position in all cases ?

client-data/tools/text/text.js
```diff
@@ -45,13 +45,16 @@
 		"lastSending": 0
 	};
 
+	var curMode = "";
```
lovasoa, line 48: Maybe this could be a boolean instead ?

client-data/board.html
finnboeger: That would probably be the way to go, do you want to go ahead and change it?

client-data/js/board.js
finnboeger: No Edge has a different UA that completely drops the MSIE monicker. Only IE<=10 had MSIE and IE 11 had Trident
`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.74 Safari/537.36 Edg/79.0.309.43`

lovasoa: Perfect :)

client-data/tools/cursor/cursor.js
finnboeger: You mean the `x` and `y` property? 
There was a performance benefit to using css transforms as it avoids content reflow and allows gpu acceleration.

client-data/tools/text/text.js
finnboeger: as it only has two modes, yes of course

client-data/board.html
lovasoa: Yes, I can do that :)

client-data/js/board.js
finnboeger: same as above then :+1:

client-data/js/canvascolor/canvascolor.min.js
finnboeger: No objection from me here

client-data/js/board.js
finnboeger: done

client-data/board.html
finnboeger: I have gone ahead and implemented it

lovasoa:
Thank you !

PR #64, opened 2020-05-04
Fix grid by moving pattern definition back into js

finnboeger:
The move of the patterns to their own file somehow broke the grid in chrome as well which slipped my initial testing.
Since I've made the effort of getting WBO ready for IE again this change is also necessary for that.

<sub>
</sub>


lovasoa:
Thank you ! I should have tested it...

PR #66, opened 2020-05-04
add option to block tools

finnboeger:
<sub>
</sub>

### DIFF
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -154,10 +154,22 @@ Tools.HTML = {
 
 Tools.list = {}; // An array of all known tools. {"toolName" : {toolObject}}
 
+Tools.setToolId = function setToolId(tool) {
+	tool._id = tool.name.split(" ").join("_").toLowerCase();
+	return tool._id;
+};
+
+Tools.isBlocked = function toolIsBanned(tool) {
+	if (!tool._id) Tools.setToolId(tool);
+	return Tools.server_config.BLOCKED_TOOLS.includes(tool._id);
+};
+
 /**
  * Register a new tool, without touching the User Interface
  */
 Tools.register = function registerTool(newTool) {
+	if (Tools.isBlocked(newTool)) return;
+
 	if (newTool.name in Tools.list) {
 		console.log("Tools.add: The tool '" + newTool.name + "' is already" +
 			"in the list. Updating it...");
@@ -182,12 +194,14 @@ Tools.register = function registerTool(newTool) {
 			newTool.draw(msg, false);
 		}
 	}
-}
+};
 
 /**
  * Add a new tool to the user interface
  */
 Tools.add = function (newTool) {
+	if (Tools.isBlocked(newTool)) return;
+
 	Tools.register(newTool);
 
 	if (newTool.stylesheet) {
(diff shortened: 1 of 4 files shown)


finnboeger:
@lovasoa you might want to cancel the CI test here, it has been running for two and a half hours for some reason.

server/sockets.js
```diff
@@ -96,6 +96,8 @@ function socketConnection(socket) {
 			return;
 		}
 
+		if (!message.data.tool ||
+			config.BLOCKED_TOOLS.includes(message.data.tool.split(" ").join("_").toLowerCase())) return;
```
lovasoa, line 100: We should probably log the issue when it happens.

client-data/js/board.js
```diff
@@ -154,10 +154,22 @@ Tools.HTML = {
 
 Tools.list = {}; // An array of all known tools. {"toolName" : {toolObject}}
 
+Tools.setToolId = function setToolId(tool) {
+	tool._id = tool.name.split(" ").join("_").toLowerCase();
```
lovasoa, line 158: It may be simpler to just comme-separate the raw list of tool names in the configuration, to avoid having to add a separate `_id` that is used only for this feature.

client-data/js/board.js
```diff
@@ -154,10 +154,22 @@ Tools.HTML = {
...
+	return tool._id;
+};
+
+Tools.isBlocked = function toolIsBanned(tool) {
+	if (!tool._id) Tools.setToolId(tool);
+	return Tools.server_config.BLOCKED_TOOLS.includes(tool._id);
+};
```
lovasoa, line 165:
```suggestion
Tools.isBlocked = function toolIsBanned(tool) {
	return Tools.server_config.BLOCKED_TOOLS.includes(tool.name);
};
```

server/configuration.js
```diff
@@ -31,4 +31,7 @@ module.exports = {
 
     /** Duration after which the emit count is reset in miliseconds */
     MAX_EMIT_COUNT_PERIOD: parseInt(process.env['WBO_MAX_EMIT_COUNT_PERIOD']) || 4096,
+
+    /** Blocked Tools. Replace spaces in the tool names with underscores and separate tool names with spaces */
+    BLOCKED_TOOLS: process.env['WBO_BLOCKED_TOOLS'] ? process.env['WBO_BLOCKED_TOOLS'].toLowerCase().split(" ") : [],
```
lovasoa, line 36:
```suggestion
    BLOCKED_TOOLS: (process.env['WBO_BLOCKED_TOOLS'] || "").split(','),
```

server/sockets.js
lovasoa, line 100:
```suggestion
		if (!message.data.tool || config.BLOCKED_TOOLS.includes(message.data.tool) {
			log('BLOCKED MESSAGE', message.data);
			return;
		}
```

finnboeger:
I applied your fixes for now. This requires that a tool does not contain a comma in its name, otherwise it can't be blocked (same issue we had with spaces before). We should at least add an assertion that there is no comma in the tool name before adding it and otherwise throw an error

server/configuration.js
```diff
@@ -31,4 +31,7 @@ module.exports = {
 
     /** Duration after which the emit count is reset in miliseconds */
     MAX_EMIT_COUNT_PERIOD: parseInt(process.env['WBO_MAX_EMIT_COUNT_PERIOD']) || 4096,
-};
\ No newline at end of file
+
+    /** Blocked Tools. Replace spaces in the tool names with underscores and separate tool names with spaces */
```
lovasoa, line 35:
```suggestion
    /** Blocked Tools. A comma-separated list of tools that should not appear on boards. */
```

finnboeger:
I would consider this ready for merging, what do you think?

lovasoa:
Thank you ! :confetti_ball:

PR #70, opened 2020-05-06
Remove extra urlencode call

### LINKED ISSUE #69: Bug: Can't preview or download boards with spaces
When running the latest master or prod branch and creating a board with a space in its name it is possible to draw on it just fine, however if one wants to download the board or render a preview the url gets encoded twice (first by the browser changing the space to `%20`, then again by the server changing it to `%2520`). 
For some reason this does not present an issue on the wbo.ophir.dev server although the same double encoding behaviour can be observed there

finnboeger:
I have not been able to find an occasion where a board that has been created properly would become inaccessible with this. Neither have I found any security issues with not urlencoding the path.

If this is a concern nevertheless we could use a small function like this
```js
function encodeIfUnencoded(str) {
	const stripped = str.split("%").join("");
	if (encodeURIComponent(stripped) === stripped) return str;
	return encodeURIComponent(str);
}
```
that would check if any characters that could be urlencoded remain (except of course for the percent sign) and only urlencode the string if they do.

Fixes #69 
<sub>
</sub>

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -91,7 +91,7 @@ function handleRequest(request, response) {
 			break;
 
 		case "download":
-			var boardName = encodeURIComponent(parts[1]),
+			var boardName = parts[1],
 				history_file = path.join(config.HISTORY_DIR, "board-" + boardName + ".json");
 			if (parts.length > 2 && /^[0-9A-Za-z.\-]+$/.test(parts[2])) {
 				history_file += '.' + parts[2] + '.bak';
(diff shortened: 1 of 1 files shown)


lovasoa:
I think the PR in its current form introduces a directory traversal vulnerability on windows (or any platform where another character than `/` can be used as path separator). We should just check that `boardName` matches `^[\w%]*$`, and reject the request if it doesn't.

finnboeger:
Do you know why the tests fail on github quite often for node 12.x only? They all pass on my local machine using 12.13.1

lovasoa:
I think this is a race condition. The tests refresh the page and then immediately assert that an element is visible, but it may take time for the element to appear.

finnboeger:
There should probably be a timeout - otherwise the test will never fail if that element is not getting drawn

lovasoa:
Thank you @finnboeger for finding this bug and fixing it !

PR #86, opened 2020-05-30
Move

paolobolzoni:
Adds two functions to move shapes on the board. The primary function moves the first shape you click on. The secondary function moves all shapes.

If one clicks and drags on a empty area the primary function start moving the first shape that contacts the cursor.

The commit also includes two icons made on the style of the other ones.

<sub>
By opening a pull request, I certify that I hold the intellectual property of the code and the icons I am submitting, and I am granting the initial authors of WBO a perpetual, worldwide, non-exclusive, royalty-free, and irrevocable license to this code.
</sub>


lovasoa:
Hello and thank you for your PR ! This is greatly appreciated, and this feature is one of the most frequently requested ones. I am going to test this and probably tweak this  before we can merge. In particular, I'd like to avoid tight coupling between tools, which limits us in our ability to add new tools and change existing ones. For this reason, it would probably be better to use a generic solution (such as an svg `transform` attribute) than to use specific knowledge about the existing tools. It would also reduce the quantity of code needed. Tell me what you think, and whether you are ready to make these changes yourself. Thank you again !

lovasoa:
Also, as a side note: WBO is compatible with old browsers and doesn't use a javascript pre-processor, so we cannot use ES6 syntax at the moment. I haven't found a lot of time to work on tooling yet. I know it's annoying, sorry 🤷

paolobolzoni:
When I started working on this I had to do some decisions. I could:

* modify each shape; this has the advantage to reduce the changes to the server side to minimum.
* add the `translate` attribute to the svg, this would require to change how the json is stored to disk and how it is loaded, besides all shapes would not really store directly their own position, but a position and a delta. On the up side, each shape would shift at the same way.
* try to reinvent how the shapes are managed in the system to have a central interface with all the common operation. This approach would be _perhaps_ cleaner, but also the most complex and it requires deep knowledge of the system.

So at the end, I went with the first. I understand your worries though. I can probably move the second, but I need few pointers how the json data is read and transformed in the svg shapes.


About the second second point, it is not annoying at all. I actually appreciate you try to keep it simple. I already pushed a replacement. To be sure, I should avoid: arrow functions, 'let', 'const', and all the class syntax around prototypes. Is there anything more?

paolobolzoni:
Now the move functions use the transform translate SVG transformation.

The code is indeed simpler, but I had to change ellipse, line, pencil, rect, and text tools to properly construct the shapes so the changes are more invasive.

client-data/tools/ellipse/ellipse.js
```diff
@@ -142,6 +142,11 @@
         shape.cy.baseVal.value = Math.round((data['y2'] + data['y']) / 2);
         shape.rx.baseVal.value = Math.abs(data['x2'] - data['x']) / 2;
         shape.ry.baseVal.value = Math.abs(data['y2'] - data['y']) / 2;
+        if (data.deltax != undefined  ||  data.deltay != undefined) {
+            var tmatrix = Tools.getTranslateMatrix(shape);
+            tmatrix.e = data.deltax ||0;
+            tmatrix.f = data.deltay ||0;
+        }
```
lovasoa, line 149: The goal of moving to a transform was to avoid having the tools being aware one of another. The transformation  code should probably not be duplicated in all tools, but should remain in the transformation tool's code.

paolobolzoni: I am not sure how you'd like it, let's focus in one shape: the rectangle.

After implementing the move using translations rectangles, in addition of `x`, `y`, `x2`, `y2`, `stroke` ..., also have a an extra coordinate containing how much they have been moved.

To properly build a rectangle I need to set up this extra coordinate, this is what the piece of code underlined does.
That piece of code is similar in all shapes, because all shapes need to be constructed using those extra coordinates, just like to set up the `x` and `y` is similar in all shapes.

However, the moving code (how the extra coordinate is computed) is inside the `moving.js` file, the new function `Tools.getTranslateMatrix` gets (or make if not there) the transformation matrix of an svg object.

lovasoa: Yes, I understand that the shape has to be created with the right translation. I am just saying that the code doing that should not be duplicated in every tool, but should instead invoked after the element has been created, without having to change the tool's code.

paolobolzoni: That would be possible if the tools store the new `deltax` and `deltay` values somewhere. On loading the move tool can reposition everything in the right place.

Is the raw json data, and in particular the `deltax` and `deltay`, stored somewhere?

lovasoa: No, unfortunately it's not. There is not real data model, and on the client side, the data is stored only in the DOM.

paolobolzoni: I had an idea I am working on now. Tell me if it is acceptable.

Basically, in the function `messageForTool` in `js/board.js` I detect if the message has an `id` and a `deltax` or a `deltay`. In the case I send a message to the mover to reposition that shape.

It pushes a bit of the mover logic in the `board.js`, but at least it will be limited in one function that would continue working even if the mover tool is disabled.

lovasoa: Yes, this sounds good to me 👍

paolobolzoni: Please have a look

lovasoa:
Ok, it looks like I don't have the rights to make changes to this PR directly, so I opened a new one. See #97

Issue #85, opened 2020-05-30
Infinite canvas


bannmann:
According to [this article](https://zapier.com/blog/best-online-whiteboard/), one of the main benefits of digital whiteboards is their infinite canvas:

> While physical whiteboards certainly have size restrictions, going digital should mean that you're no longer limited in how big or complex your visualizations are.

However, unlike some commercial offerings, WBO does not offer this feature: When I create a new whiteboard, my viewing area is positioned on the top-left corner of the canvas. If I start drawing immediately and later realize that I want to draw something to the left or above of my initial drawing, I have no way of moving my viewing area further to the left.

Ideally, WBO would instead work like this: on a new whiteboard, the viewing area is positioned on the "center" of an effectively infinite canvas. If I use the hand tool to pan the viewing area to the top, I can draw there, and I could repeat that process endlessly.

----
_Note: if WBO offered a way to move objects, one could work around the lack of an unlimited canvas by selecting the entire drawing and moving it away from the edge of the canvas. But that would be cumbersome, especially if one has to do it repeatedly._

lordneeko:
If this does not get implemented, recommend at least have a border at the "edge" of the canvas. Right now, you can zoom WAY out, but there is no indication that the canvas is limited
![image](https://user-images.githubusercontent.com/11577533/99694627-a2b31280-2a5a-11eb-8ce8-b29d6a67141f.png)

lovasoa:
@lordneeko I think you are talking about a different issue. This one is about being able to scroll to the left and to the top of the initial position.

lordneeko:
It is relevant. The OP mentioned that there is "not" an infinite canvas...but the UI doesn't make that obvious. I get that the OP is talking about starting in the "center" versus top-left. I can just post it as a separate issue though

PR #97, opened 2020-06-21
Move

### LINKED ISSUE #10: Possibility to edit / move entities

lovasoa:
Add an element mover
Original PR by @paolobolzoni  (see #86)

Fixes #10


lovasoa:
I removed the move_all functionnality, because I'd rather avoid features that let a single user have an effect to the whole board. A single misbehaving user could prevent everyone else from working constructively. 
However, I do understand that the feature may be useful. I think implementing #85 would solve the root problem.

paolobolzoni:
Of course I am well aware that it is your project and your is the final word, but I have few comments.

Indeed to have an unlimited canvas (ie, #85) would solve the root problem better than move-all, but wouldn't be better to remove move-all after #85 is implemented? It is really annoying to start a drawing in the top right and notice you run out of space after a while for you need to add something on the right or on the top.

In my implementation the tool moves the first shape the cursor touches clicking or after while dragging. So a user does not need to do a pixel-perfect click to move small objects like thin lines or dots. The current implementation is simpler, but in this case I think a bit more implementation complexity was a good exchange for user experience: click and if you missed just drag a bit until the mover picks up the intended shape.

In the mover.js stopMoving function you just call move. However the move function broadcasts the shape position only every 70 milliseconds. Are you sure the final position is sent? Perhaps one needs to reset last_sent to 0 or something similar?

lovasoa:
Hello, and thank you for your feedback. These are only good points you are raising. To answer them:
  - A move-all tool would let a single bad actor waste the experience of all the other connected users. So I thought it would be better not to have it, even before #85. 
 - When I noticed this *hover to select* behavior, I thought it was a bug. I think many users would have taken it as a bug too. This behavior is uncommon in drawing apps, and not very discoverable... But if there is demand, we'll reintroduce it! In the meantime, you can just zoom on a small element to select it. 
 - About the 70ms threshold in release: good point. Did you notice any issue with it? We've had the same kind of issues in other tools, and I think we should handle the throttling in board.js.

paolobolzoni:
That's a fair point. I used the board only with my friends, but rogue users probably do exists in other scenarios. Still, is it really so much worse of someone dragging around the Eraser?

The hover to select was indeed intentional, but I guess the commit message where I wrote "_The mover not detect the object to move_" instead of "_The mover no**w** detect**s** the object to move_" has been very confusing (I was probably sleepy).
Still, I hardly believe it worsen the user experience, it is also hard to compare with other drawing applications as whitebophir has no concept of selection. My limited testing with friends did not bring anyone confused, they simply did not notice the possibility as they naturally tried to click on the shapes, but seldom pick up the correct shape after starting moving. I even received a positive feedback once one understood how it worked. Still, as I said, it has been quite limited testing.
About other drawing apps, I think to press the middle button should automatically use the Hand tool as in Qcad, Gimp, Inkscape, and probably many others.

About the 70ms, I did not test your version deeply, but when I was implementing I noticed it could happen that the user moving and all other users end up with shapes in different positions. It could also happen that reloading would move shapes slightly (as the server never got the final position).
The solution is pretty easy, though. Ensure to send the final position when the user stop moving. Since all positions are absolute, there cannot be confusion. Even if two users start moving the same object at the same time, only one position will "win."

Issue #109, opened 2020-08-06
New tool: mathematics


dazsmith:
Firstly, thanks for a great piece of software! I am planning to use it in my mathematics classes, where not all students have access to a pen. I want to add a tool to accept TeX and display it, based on the existing Text tool. I am using mathjax to convert the TeX to an svg, which is then added to the canvas. I had no javascript or svg experience before this week, so please excuse my novice approach.

I am working in the branch https://github.com/dazsmith/whitebophir/tree/optimize-for-pen of my fork. As you can see at https://wb.dasmithmaths.com, the mathjax calls work, and the new tool works to display typeset mathematics (try typing something like `\sqrt{x^2+1}-x^x/3`). It will even sync from one user to another, and save in the `board-anonymous.json` file. But, when loading the page (after a refresh, for example), the `svg` elements that should contain the mathematics are empty. They are also missing their width, height and viewBox attributes. I also can't see how to enable editing of an existing mathematics element.

If you can offer any assistance, that would be really great! Sorry to be begging for assistance, but I am getting stuck.

dazsmith:
I figured out the display problem. I was failing to call `updateMathematics` after `createMathematicsField`. I'll keep working on the selection / editing problem.

dazsmith:
Selection & editing now works. The eraser does not work on these objects, yet. It also would be nice to make the typeset mathematics default to the same color as the user selected.

dazsmith:
Moving, erasing, and automatic color now works.

I also added a tool to insert image from URL. The latter is a bit clunky because I don't understand promises. If you paste in the URL of a valid image, then the image should appear. At the moment, you have to press `[space]` a few times while waiting for the browser to cache the image, then it will appear. The image is automatically immune from the eraser; to change that, edit `INDELIBLE` to `DELIBLE` before the image URL. Check it out at https://wb.dasmithmaths.com/.

Should I submit a pull request?

lovasoa:
Hello ! This is a nice idea ! Yes, please open a pull request for the math tool.

ClaudeStabile:
@lovasoa @dazsmith Is there a docker image ready to use for this wbo version ? What is the pull to get it ?
I am running a wbo instance @ https://wbo.free-solutions.org , it is also integrated with our vidconf/chat services @ https://www.free-solutions.org Math option and file upload is a nice addon very needed by classroom in the COVID times

Congrats to both efforts and innovations. I love WBO as all Cobrowsing enabled apps

[...]

PR #124, opened 2020-08-24
Added mathematics tool.

dazsmith:
<sub>
</sub>

This code adds a "mathematics tool" which accepts (La)TeX code in a text box, formats it using mathjax, and displays it on the page. This is designed to address #109.

### Explanation of code
- The input LaTeX is stored in the `aria-label` attribute of the svg element created by mathjax. I don't think this is contrary to the designated use of this accessibility attribute, as it is reasonable to expect that a screen reader would benefit from accessing the input LaTeX code.
- The color of the displayed mathematics is the color selected by the user, but the usual LaTeX color commands can be used to (locally) change the color. The user's color is stored in the `stroke` attribute of the rectangle of class `clickhelper`.
- The rectangle of class `clickhelper` is appended to the output of mathjax to make it easier to select the displayed mathematics for future edits & moves using the move tool.
- The edits to ellipse, pencil and rectangle tools and associated CSS and board.css are because mathjax outputs an svg element containing many other elements, whose `fill` attribute should *not* be set to `none`. Therefore, a new class, `nofill`, was created for those tools providing the original functionality.
- The edits to the hand tool are to enable selection of the "parent" svg element containing all the mathjax output, so that not just a single part of the mathjax output is moved.

### Some comments
- I don't know whether or not line 7 of the new mathematics.js should be edited to include my name too. Please advise.


dazsmith:
Perhaps the failed checks failed because of the edits to the pencil tool?

dazsmith:
It was not the change to the pencil tool.
The issue was that the main canvas svg node required property `xmlns:xlink="http://www.w3.org/1999/xlink"` to be able to contain output of mathematics tool.
Solving that problem revealed a typo in a line I added.

I'm not sure why this issue only arises in the `preview` and not the `boards`.

droundy:
This sounds great. Would you guess that this pull request is in a state where I could reasonably integrate it intro my whiteboards for classes starting in under two weeks?

dazsmith:
I've been using it in my classes for 5 weeks with no problem.

droundy:
If the mathjax a standard 2.7, or did you have to customize it? If it's standard, is there any reason not to source it from a CDN?

droundy:
When I merge this with my code, I'm getting a problem in the preview, which unfortunately is pretty high priority for my plans to manage my dozen breakout rooms.  :(

droundy:
I'm not sure how to directly contribute to someone else's pull request, so I'll note the one change I had to make here:
```
$ git diff
diff --git a/client-data/view.html b/client-data/view.html
index f595660..a22ff21 100644
--- a/client-data/view.html
+++ b/client-data/view.html
@@ -21,6 +21,7 @@
        <link rel="alternate" hreflang="{{.}}" href="{{../boardUriComponent}}?lang={{.}}" />
        {{/languages}}
        <script src="../polyfill.min.js"></script>
+       <script type="text/javascript" id="MathJax-script" src="https://mj.dasmithmaths.com/mathjax/tex-svg-full.js"></script>
 </head>
 
 <body>
@@ -94,6 +95,7 @@
        <script src="../tools/rect/rect.js"></script>
        <script src="../tools/ellipse/ellipse.js"></script>
        <script src="../tools/text/text.js"></script>
+       <script src="../tools/mathematics/mathematics.js"></script>
        <script src="../tools/eraser/eraser.js"></script>
        <script src="../tools/hand/hand.js"></script>
        <script src="../tools/grid/grid.js"></script>
```

Thanks for this great tool!

droundy:
I notice that when I use math like $\frac{\sin x}{\cos x}$ the rendered math is often hidden by the edit field.  Would it be possible to shift the edit field down just a bit more to help with this? It's even worse if I use an align environment or similar, so maybe we could just make the edit a bit transparent?

dazsmith:
> If the mathjax a standard 2.7, or did you have to customize it? If it's standard, is there any reason not to source it from a CDN?

It is standard mathjax, called with `<script type="text/javascript" id="MathJax-script" src="https://mj.dasmithmaths.com/mathjax/tex-svg-full.js"></script>`. I have never used mathjax before, so did not know if I could trust a public server, so just rolled my own. I haven't tried it with a public server, but I don't see why it wouldn't work, as long as they don't rate limit the mathjax requests in any way.

dazsmith:
> I'm not sure how to directly contribute to someone else's pull request, so I'll note the one change I had to make here:
> ...
> Thanks for this great tool!

Thanks for the contribution. I can't find `/client-data/view.html` in either of lovasoa's branches. Where is this file?

Once I understand that ... I also don't know the best way for you to edit a pr, but I guess you could submit a pr to my repository, which I accept, thereby updating the branch for this pr too.

dazsmith:
> I notice that when I use math like $\frac{\sin x}{\cos x}$ the rendered math is often hidden by the edit field. Would it be possible to shift the edit field down just a bit more to help with this? It's even worse if I use an align environment or similar, so maybe we could just make the edit a bit transparent?

I had thought of trying to move the textbox *up* a bit but couldn't quickly figure it out so gave up. Transparency is a much better idea. I'll give it a go in the coming week.

droundy:
Apparently I didn't make a pull request that I thought I had. At least I can't find record of it.  So it seems that `view.html` is only in my repository.  I should put together a pull request...

droundy:
Hi @dazsmith , I just sent a pull request to you with some useful changes, hopefully you can merge it into this just fine.

dazsmith:
I'm not sure why some of the checks failed and others passed. Can you comment, @lovasoa or @droundy ?

sents:
Note that this PR uses setting of `innerHTML` of the data sent to the server. This allows potentially setting arbitrary HTML in the svg element. Setting `innerHTML` in an `<svg>` environment doesn't do anything so it doesn't allow arbitrary code to be executed to everybody seeing the board. However if somebody were to download the svg and embed in in a page a script tag would be executed. If there is interest in merging this eventually I can add code to filter `<script>`, `<image>`, `<a>` on the server side.

sents:
@lovasoa I would really like to work on this PR but I can not commit to this pull request. Should I open another pull request?
@dazsmith Are you still interested on working on this?

dazsmith:
@sents I would like to come back to working on this. I think if you submit a pull request to me, then I can merge it into this one.
I expect that with the upstream updates there is quite a bit of work to be done.
I'm not sure why it always failed the checks, but my guess is that downloading mathjax takes so long that it causes a timeout.

Your plan to filter `<script>` etc sounds good.

holema:
Will you merge this PR @lovasoa? This PR seems to be great

PR #137, opened 2020-09-24
environment variable WBO_MAX_BOARD_SIZE_X to give max width of board (closes #118)

### LINKED ISSUE #118: Configuration to set the width of the board?
It would be lovely to have a configuration to set the width of the board, and just have that fixed.  That would force users to only scroll down, which I think would make it harder to simply fail to notice that something was written off-screen.

droundy:
I just added a configuration variable MAX_BOARD_SIZE_X.  I also made `board.js` obey the existing `MAX_BOARD_SIZE` (which it wasn't), and made it so you can't zoom out beyond the width of the board.  The latter isn't technically necessary, and could in cases of a very wide window be annoying, but I think it's better than letting users confuse themselves with a bunch of white space that they can't write on.

<sub>
</sub>

### DIFF
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -418,7 +418,7 @@ function resizeCanvas(m) {
         Tools.svg.width.baseVal.value = Math.min(x + 2000, Tools.server_config.MAX_BOARD_SIZE_X);
 	}
 	if (y > Tools.svg.height.baseVal.value - 2000) {
-        Tools.svg.height.baseVal.value = Math.min(y + 2000, Tools.server_config.MAX_BOARD_SIZE);
+		Tools.svg.height.baseVal.value = Math.min(y + 2000, Tools.server_config.MAX_BOARD_SIZE);
 	}
 }
 
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -414,12 +414,11 @@ function updateDocumentTitle() {
 function resizeCanvas(m) {
 	//Enlarge the canvas whenever something is drawn near its border
 	var x = m.x | 0, y = m.y | 0
-	var MAX_BOARD_SIZE = 65536; // Maximum value for any x or y on the board
 	if (x > Tools.svg.width.baseVal.value - 2000) {
-		Tools.svg.width.baseVal.value = Math.min(x + 2000, MAX_BOARD_SIZE);
+        Tools.svg.width.baseVal.value = Math.min(x + 2000, Tools.server_config.MAX_BOARD_SIZE_X);
 	}
 	if (y > Tools.svg.height.baseVal.value - 2000) {
-		Tools.svg.height.baseVal.value = Math.min(y + 2000, MAX_BOARD_SIZE);
+        Tools.svg.height.baseVal.value = Math.min(y + 2000, Tools.server_config.MAX_BOARD_SIZE);
 	}
 }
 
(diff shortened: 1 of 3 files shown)


client-data/js/board.js
```diff
@@ -436,6 +435,7 @@ var scaleTimeout = null;
 Tools.setScale = function setScale(scale) {
 	if (isNaN(scale)) scale = 1;
 	scale = Math.max(0.1, Math.min(10, scale));
+	scale = Math.min(scale, document.documentElement.clientWidth/Tools.server_config.MAX_BOARD_SIZE_X);
```
lovasoa, line 438: We could just add this to the Math.min above

```suggestion
	var max_scale = document.documentElement.clientWidth/Tools.server_config.MAX_BOARD_SIZE_X;
	scale = Math.max(0.1, Math.min(10, max_scale, scale));
```

client-data/js/board.js
```diff
@@ -414,12 +414,11 @@ function updateDocumentTitle() {
...
 	//Enlarge the canvas whenever something is drawn near its border
 	var x = m.x | 0, y = m.y | 0
-	var MAX_BOARD_SIZE = 65536; // Maximum value for any x or y on the board
 	if (x > Tools.svg.width.baseVal.value - 2000) {
-		Tools.svg.width.baseVal.value = Math.min(x + 2000, MAX_BOARD_SIZE);
+        Tools.svg.width.baseVal.value = Math.min(x + 2000, Tools.server_config.MAX_BOARD_SIZE_X);
 	}
 	if (y > Tools.svg.height.baseVal.value - 2000) {
-		Tools.svg.height.baseVal.value = Math.min(y + 2000, MAX_BOARD_SIZE);
+        Tools.svg.height.baseVal.value = Math.min(y + 2000, Tools.server_config.MAX_BOARD_SIZE);
```
lovasoa, line 421: This mixes tabs and spaces

```suggestion
		Tools.svg.height.baseVal.value = Math.min(y + 2000, Tools.server_config.MAX_BOARD_SIZE);
```

client-data/js/board.js
```diff
@@ -414,12 +414,11 @@ function updateDocumentTitle() {
 function resizeCanvas(m) {
 	//Enlarge the canvas whenever something is drawn near its border
 	var x = m.x | 0, y = m.y | 0
-	var MAX_BOARD_SIZE = 65536; // Maximum value for any x or y on the board
 	if (x > Tools.svg.width.baseVal.value - 2000) {
-		Tools.svg.width.baseVal.value = Math.min(x + 2000, MAX_BOARD_SIZE);
+        Tools.svg.width.baseVal.value = Math.min(x + 2000, Tools.server_config.MAX_BOARD_SIZE_X);
```
lovasoa, line 418:
```suggestion
		Tools.svg.width.baseVal.value = Math.min(x + 2000, Tools.server_config.MAX_BOARD_SIZE_X);
```

lovasoa (changes requested):
This looks good, I'm ready to merge that. I agree that getting rid of the uneditable blank spaces is worth it, even if it is an inconvenience on screens with an unusual aspect ratio.

lovasoa:
Can you have a look at the failing tests too ?

droundy:
It looks like the failing tests assume a given transformation matrix, which is getting broken by my code that sets the minimum zoom level.

lovasoa:
It looks like the cursor element (to which the transform should be applied) is not even present

Issue #151, opened 2020-11-25
Button for clearing a whiteboard


mwllgr:
Similar to #120 - why isn't there a button to clear the whole board? It's easier than deleting the file or sharing a new link.

topninja:
hello @mwllgr 
I tried to make myself.

/client-data/board.html
insert this code under any icon what you want.
if you want to lay this icon at top of menu, then add under `<script src="../tools/pencil/wbo_pencil_point.js"></script>`
`<script src="../tools/new/new.js"></script>`

create "new" folder inside /tools
inside this directory, make new file as new.js, new.svg(add any icon what you want)
this is new.js file content.
```

(function grid() { //Code isolation

    function newPanel(evt) {
        Tools.drawingArea.innerHTML = ''
    }

    Tools.add({ //The new tool
        "name": "New",
        "shortcut": "n",
        "listeners": {},
        "icon": "tools/new/new.svg",
        "oneTouch": true,
        "onstart": newPanel,
        "mouseCursor": "crosshair",
    });

})(); //End of code isolation
```

Thanks

mwllgr:
Thanks @topninja! Any info on how I can integrate this with the Docker container?

lovasoa:
> 
> ```
> 
> (function grid() { //Code isolation
> 
>     function newPanel(evt) {
>         Tools.drawingArea.innerHTML = ''
>     }
> 
>     Tools.add({ //The new tool
>         "name": "New",
>         "shortcut": "n",
>         "listeners": {},
>         "icon": "tools/new/new.svg",
>         "oneTouch": true,
>         "onstart": newPanel,
>         "mouseCursor": "crosshair",
>     });
> 
> })(); //End of code isolation
> ```

This does not clear the board. This just clears the drawing area locally on your own screen, without sending anything to the server or to other connected users. If you simply refresh the page, you will the the board reappear.

topninja:
sorry for check in deep.
I revised again. I checked this is working as well.

revise above new.js like this
```



(function grid() {
    
    function newPanel() {
        Tools.socket.emit('broadcast', {
            board: Tools.boardName,
            data : {
                tool : "New",
                type : "deleteall"
            }
        });
        Tools.drawingArea.innerHTML = '';
    }

    function eraseAll(data) {
		switch (data.type) {
			case "deleteall":
                Tools.drawingArea.innerHTML = '';
				break;
		}
    }

    Tools.add({
        "name": "New",
        "shortcut": "n",
[...]

Issue #120, opened 2020-08-21
How to clear a whiteboard?


droundy:
I'd like to have some way on the server to clear a set of whiteboards.  I'm finding that content insistently returns, even after I delete the json files from server-data and restart the npm process.  Is there some other trick I need?

mwllgr:
Sorry for reopening this - but a simple button to clear all changes would be a nice feature too.

Issue #92, opened 2020-06-19
Configurable 'clear' button and canvas colour.


pandian-egnaroinc:
Hi,
Continuing the discussion from https://github.com/jitsi/jitsi-meet/issues/5295

A clear button disabled by default and enabled based on a configuration variable will help self-hosted server admins to make the decision whether to enable of not. Same is the case with canvas colours to choose whiteboard vs black board etc.

I agree to the rationale of being hesitant to make it default behaviour. Neither am I asking to change the default behaviour of https://wbo.ophir.dev/

lovasoa:
Related to #90

jondo:
@pandian-egnaroinc, I think the configurable canvas colour should be filed as separate issue.

pandian-egnaroinc:
> @pandian-egnaroinc, I think the configurable canvas colour should be filed as separate issue.

Sure, Will file a new issue.

Thanks,

danionescu2007:
Hello, 

Is there a way to clear a white board, by user or by closing it, yet?
Thank you.

lovasoa:
Hello @danionescu2007 ! No there isn't.

danionescu2007:
Thank you for the info. Too bad 😀 there isn't.

lovasoa:
This issue now splits into already-handled and still-open parts.

The clear button exists as the Clear tool. It is registered as `toolId: "clear"` in `client-data/tools/manifest.js`, implemented in `client-data/tools/clear/index.js`, and gated by the `clearBoard` capability so deployments can distinguish ordinary edit permission from board-wide clear permission.

The configurable canvas color / blackboard / transparent background part is still open and tracked more directly in #96.

Closing this as superseded: clear is implemented, and the remaining canvas-color work should continue in #96.

Issue #90, opened 2020-06-15
how to clear the whiteboard on close


sarathic07:

lovasoa:
This request is not the same as the Clear tool: it asks for automatic cleanup when a board/browser closes.

That behavior is better tracked by #199 ("Delete Board as the last user leaves"), because implementing it safely needs board/session lifecycle semantics rather than just a UI button. It needs to define what "last user" means, how reconnects are handled, whether the persisted board file is deleted, and how to avoid data loss during transient disconnects.

Closing this older empty issue as a duplicate of #199.

PR #152, opened 2020-11-27
Erase all button on menu

topninja:
this is for #151  #120 
add erase all button on menu

### DIFF
--- a/client-data/board.html
+++ b/client-data/board.html
@@ -90,6 +90,7 @@
 	<script src="../js/minitpl.js"></script>
 	<script src="../js/board.js"></script>
 	<script src="../tools/pencil/wbo_pencil_point.js"></script>
+	<script src="../tools/new/new.js"></script>
 	<script src="../tools/pencil/pencil.js"></script>
 	<script src="../tools/cursor/cursor.js"></script>
 	<script src="../tools/line/line.js"></script>
--- a/client-data/tools/new/new.js
+++ b/client-data/tools/new/new.js
@@ -1,0 +1,60 @@
+/**
+ *                        WHITEBOPHIR
+ *********************************************************
+ * @licstart  The following is the entire license notice for the 
+ *  JavaScript code in this page.
+ *
+ * Copyright (C) 2013  Ophir LOJKINE
+ *
+ *
+ * The JavaScript code in this page is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU
+ * General Public License (GNU GPL) as published by the Free Software
+ * Foundation, either version 3 of the License, or (at your option)
+ * any later version.  The code is distributed WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU GPL for more details.
+ *
+ * As additional permission under GNU GPL version 3 section 7, you
+ * may distribute non-source (e.g., minimized or compacted) forms of
+ * that code without the copy of the GNU GPL normally required by
+ * section 4, provided you include this license notice and a URL
+ * through which recipients can access the Corresponding Source.
+ *
+ * @licend
+ */
+
+(function grid() {
+    // when new button clicks then calls this function
+    function btnClickHandler() {
+        Tools.socket.emit('broadcast', {
+            board: Tools.boardName,
+            data : {
+                tool : "New",
+                type : "deleteall"
+            }
+        });
+        Tools.drawingArea.innerHTML = '';
+    }
+    
+    // this is callback handler from socket
+    function eraseAll(data) {
+		switch (data.type) {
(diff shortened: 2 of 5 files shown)


client-data/board.html
```diff
@@ -90,6 +90,7 @@
 	<script src="../js/minitpl.js"></script>
 	<script src="../js/board.js"></script>
 	<script src="../tools/pencil/wbo_pencil_point.js"></script>
+	<script src="../tools/new/new.js"></script>
```
lovasoa, line 93: It should probably not be called new, but something like clear

client-data/tools/new/new.js
```diff
@@ -0,0 +1,60 @@
...
+    // when new button clicks then calls this function
+    function btnClickHandler() {
+        Tools.socket.emit('broadcast', {
+            board: Tools.boardName,
+            data : {
+                tool : "New",
+                type : "deleteall"
+            }
+        });
+        Tools.drawingArea.innerHTML = '';
```
lovasoa, line 37: Tools.socket shouldn't be used directly. You can use drawAndSend. See other tools for an example.

client-data/tools/new/new.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+    
+    // this is callback handler from socket
+    function eraseAll(data) {
+		switch (data.type) {
+			case "deleteall":
+                Tools.drawingArea.innerHTML = '';
+				break;
```
lovasoa, line 45: You have a formatting problem

client-data/tools/new/new.js
```diff
@@ -0,0 +1,60 @@
...
+		}
+    }
+
+    Tools.add({
+        "name": "New",
```
lovasoa, line 50: it shouldn't be called new

client-data/tools/new/new.js
```diff
@@ -0,0 +1,60 @@
...
+        "shortcut": "n",
+        "listeners": {},
+        "icon": "tools/new/new.svg",
```
lovasoa, line 53: not new

client-data/tools/new/new.js
lovasoa: Also, there should probably be a confirmation prompt. This erases all data forever, which could represent days of work by multiple people.

client-data/tools/new/new.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 24.1.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 512 512" style="enable-background:new 0 0 512 512;" xml:space="preserve">
+<g>
+	<polygon points="448,80 512,80 512,112 448,112 448,176 416,176 416,112 352,112 352,80 416,80 416,16 448,16 	"/>
+	<path d="M32,112h256V80H32C14.3,80,0,94.3,0,112v352c0,17.7,14.3,32,32,32h384c17.7,0,32-14.3,32-32V247.3h-32V464H32V112z"/>
+</g>
+</svg>
```
lovasoa, line 9: This icon makes it look like something new is being created, wheras all the data on the board is erased and lost forever.

server/boardData.js
```diff
@@ -97,6 +97,14 @@ BoardData.prototype.delete = function (id) {
 	this.delaySave();
 };
 
+/** Removes all data from the board
+ */
+BoardData.prototype.deleteall = function () {
+	//KISS
+	this.board = [];
```
lovasoa, line 104: This.board is an object, not an array

server/sockets.js
```diff
@@ -143,6 +143,9 @@ async function saveHistory(boardName, message) {
 		case "delete":
 			if (id) board.delete(id);
 			break;
+		case "deleteall":
+			board.deleteall();
```
lovasoa, line 147: We should check that the tool is activated, and this tool should probably not be activated by default.

lovasoa (changes requested):

lovasoa:
See https://github.com/lovasoa/whitebophir/issues/92 and https://github.com/lovasoa/whitebophir/issues/90 and the discussion in https://github.com/jitsi/jitsi-meet/issues/5295

Such a tool is very dangerous, since it allows a single misbehaving user to sabotage the work of all others on a large whiteboard. This tool should be disabled by default, and be possible to enable through configuration.

topninja:
I agree with your suggesstion @lovasoa

willerhe (approved):

willerhe:
Sorry About that,I recklessly clicked APPROVED button.

PR #166, opened 2021-02-06
Remember and restore white-out size

ishabalin:
Remember "White-Out" and "Pencil" sizes and restore when switching between the two. This is helpful when using "White-Out" as an eraser and switching frequently, as one typically wants eraser to have larger radius.

<sub>
</sub>

### DIFF
--- a/client-data/tools/pencil/pencil.js
+++ b/client-data/tools/pencil/pencil.js
@@ -139,6 +139,24 @@
 		return line;
 	}
 
+	//Remember primary and secondary sizes separately
+	var pencilSize = -1;
+	var whiteOutSize = -1;
+
+	//Restore remembered size after switch
+	function toggleSize() {
+		if (pencilTool.secondary.active) {
+			pencilSize = Tools.getSize();
+			if (whiteOutSize != -1) {
+				Tools.setSize(whiteOutSize);
+			}
+		} else {
+			whiteOutSize = Tools.getSize();
+			if (pencilSize != -1) {
+				Tools.setSize(pencilSize);
+			}
+		}
+	}
 
 	var pencilTool = {
 		"name": "Pencil",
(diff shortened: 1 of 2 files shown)


client-data/tools/pencil/pencil.js
```diff
@@ -154,10 +154,12 @@
 			"icon": "tools/pencil/whiteout_tape.svg",
 			"active": false,
 			"switch": stopLine,
+			"size": -1
 		},
 		"mouseCursor": "url('tools/pencil/cursor.svg'), crosshair",
 		"icon": "tools/pencil/icon.svg",
-		"stylesheet": "tools/pencil/pencil.css"
+		"stylesheet": "tools/pencil/pencil.css",
+		"size": -1
```
lovasoa, line 162: Having to specify `size: -1` in two distinct places in a tool to give it this ability is a little confusing. You should probably use the existing `secondary.switch` function to implement that.

lovasoa (changes requested):
The feature is nice and I am willing to merge it, but the implementation is problematic.

client-data/tools/pencil/pencil.js
ishabalin: Fair point. One of the ideas behind two separate `size` attributes was one can specify different default sizes. I will look into more elegant implementation.

client-data/tools/pencil/pencil.js
```diff
@@ -153,11 +171,20 @@
...
 			"icon": "tools/pencil/whiteout_tape.svg",
 			"active": false,
-			"switch": stopLine,
+			"switch": function() {
+				stopLine();
+				toggleSize();
+			},
+		},
+		"onquit": function() {
+			//When switching to another tool, restore Pencil and its size
```
ishabalin, line 180: Not sure though, if this would conflict with the way other people may use White-out?

client-data/tools/pencil/pencil.js
```diff
@@ -153,11 +171,20 @@
...
+			if (pencilTool.secondary.active) {
+				Tools.change("Pencil");
+			}
 		},
```
lovasoa, line 184: Yes, I don't think we want to switch back to pencil when changing tools. People can do it manually if they want; but I think automatically switching back to primary only for the pencil tool is confusing.

```suggestion
```

lovasoa (changes requested):

client-data/tools/pencil/pencil.js
ishabalin: Agreed, that was confusing. But here's what I wanted to achieve by that. Imagine I'm using a Pencil, then switch to White-out, set different size, correct my drawing, then switch to Rectangle. Now my expectation Rectangle would have the same line width Pencil had, but instead it inherits White-out size. In the latest commit I'm addressing that by treating Pencil line width and all other tool line widths as "Drawing" line width, and White-out has its individual line width.

lovasoa: I don't think it's surprising that WBO **doesn't change** the line width when you switch between tools.
We may want to have a separate width for each tool, but this is a different question.

lovasoa (approved):
Ok, let's merge it like that, and see how usable it is that way.

PR #167, opened 2021-02-06
Auto white-out with a finger when using stylus

ishabalin:
It's handy to be able to use a finger to white-out while using a stylus for drawing, just as you would on a physical whiteboard.

<sub>
</sub>


lovasoa:
I cannot find any documentation on `touchType`, and it looks like it is not a standard, but a custom apple extension that works only on mobile safari... Isn't there a standard way to handle that ?

ishabalin:
> I cannot find any documentation on `touchType`, and it looks like it is not a standard, but a custom apple extension that works only on mobile safari... Isn't there a standard way to handle that ?

I wish there was, so far there is only the draft https://w3c.github.io/touch-events/

lovasoa:
Ok, let's keep it like that, then.

client-data/tools/pencil/pencil.js
```diff
@@ -149,6 +168,11 @@
 			"release": stopLineAt,
 		},
 		"draw": draw,
+		"stylus": false,
```
lovasoa, line 171: This doesn't need to be globally available outside this tool if this is meant to be used only here, it can be a local variable.
If we want to be able to reuse this in other tools, then this should be `Tools.has_used_stylus` instead.

lovasoa (changes requested):

client-data/tools/pencil/pencil.js
ishabalin: Haven't thought about how this might work for other tools, so changed to a local variable for now.

server/configuration.js
```diff
@@ -1,6 +1,13 @@
 const path = require("path");
 const app_root = path.dirname(__dirname); // Parent of the directory where this file is
 
+function parseBool(val) {
+    switch(val && val.toLowerCase && val.toLowerCase().trim()) {
+        case "true": case "yes": case "y": case "1": return true;
+        default: return false;
+    }
+}
+
```
lovasoa, line 10: I don't think we need that.

server/configuration.js
```diff
@@ -39,4 +46,7 @@ module.exports = {
 
     /** Blocked Tools. A comma-separated list of tools that should not appear on boards. */
     BLOCKED_TOOLS: (process.env['WBO_BLOCKED_TOOLS'] || "").split(','),
+
+    /** Automatically switch to White-out on finger touch after drawing with Pencil using a stylus */
+    AUTO_FINGER_WHITEOUT: parseBool(process.env['AUTO_FINGER_WHITEOUT']),
```
lovasoa, line 51: Enabling it won't have any impact on any device but the iPad, right ? And you say it's the most intuitive on the iPad ? If so, I think we can enable it by default.

```suggestion
    AUTO_FINGER_WHITEOUT: process.env['AUTO_FINGER_WHITEOUT'] !== "disabled",
```

Very few people change the defaults.

server/configuration.js
```diff
@@ -39,4 +46,7 @@ module.exports = {
 
     /** Blocked Tools. A comma-separated list of tools that should not appear on boards. */
     BLOCKED_TOOLS: (process.env['WBO_BLOCKED_TOOLS'] || "").split(','),
+
+    /** Automatically switch to White-out on finger touch after drawing with Pencil using a stylus */
```
lovasoa, line 50: You should specify that this setting impacts only iPad users.

lovasoa (changes requested):

lovasoa (approved):
Thank you for your contribution !

PR #169, opened 2021-02-12
Docker - Remove unnecessary permission change of bind-mount directory

ilkohoffmann:
While using the Docker-Image I stumbled upon the required permission change to the bind-mount directory in order to persist the boards outside of docker. This is not neccessary imo. I have updated the Dockerfile and the corresponding section in the Readme. Tested on OSX 10.14.6 and Ubuntu 18.04.5.
<sub>
</sub>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,6 @@
 FROM node:14-alpine
 
-
 WORKDIR /opt/app
-
-RUN chown -R 1000:1000 /opt/app
-
-USER 1000:1000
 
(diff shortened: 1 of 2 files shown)


lovasoa:
The change was made on purpose as a security measure. Running as root is an (unfortunately common) bad practice.

> If a service can run without privileges, use USER to change to a non-root user.
> [*Docker official documentation*](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user)

Issue #168, opened 2021-02-12
Board navigation


timconsidine:
Great app !
But :

- how to save / exit a board ?  seems only option is to close browser window
- how to open a previous board ? seems only way is to reload url

I think I can cope with this. but from user perspective, closing the window seems like throwing their work.

And re-opening would be nicer with a menu of boards.  Or is this a user authentication issue ?

lovasoa:
Hello !

> how to save / exit a board ? seems only option is to close browser window

Yes, the board is always saved, every change is immediately persisted. You can close it when you want and you won't lose anything.

> how to open a previous board ? seems only way is to reload url

You can also type the board name in the text field on the main page.

<img width="1284" alt="image" src="https://user-images.githubusercontent.com/552629/107876297-7bd37380-6ec5-11eb-93be-e0ae04f8faa2.png">

nneka-nu:
Hello, @lovasoa. I'd like to work on this issue.

I figured I could save board names to localStorage in `board.js`. But since localStorage is not available on the server, I can’t think of a way to pass the data to the home page. Any ideas?

What if instead of using localStorage, I save the most recent 20 board names in a file called `recent-boards.json`, which will be stored in the `server-data` directory?

lovasoa:
Hello !

> I'd like to work on this issue.

Great to hear ! 

> I figured I could save board names to localStorage in board.js. But since localStorage is not available on the server, I can’t think of a way to pass the data to the home page. Any ideas?

Yes, the personal board history should not be saved on the server. It could be saved in localStorage, and rendered dynamically on the client-side. You  will have to add a script to [the index page](https://github.com/lovasoa/whitebophir/blob/master/client-data/index.html) to render the list.

nneka-nu:
Ah, of course! Thank you!

PR #183, opened 2021-04-01
Show a list of recent boards on the home page

nneka-nu:
Completes #168.
<sub>
</sub>


server/translations.json
```diff
@@ -25,6 +25,7 @@
     "pencil": "Stift",
     "private_board_description": "Du kannst ein <b>privates Whiteboard</b> mit einem zufälligen Namen erstellen, welches man nur mit seinem Link öffnen kann. Benutze dies wenn du private Informationen teilen möchtest.",
     "public_board_description": " Das <b>öffentliche Whiteboard</b> kann von jedem geöffnet werden. Es ein fröhliches unorganisiertes Chaos wo du zusammen mit anonymen Fremden malen kannst. Alles dort ist vergänglich.",
+    "recent_boards": "Neueste Boards",
```
lovasoa, line 28: Where did you get these translations ? Do you speak all of these languages ? If not, then please be very careful with the translations, and just put the string in english in case of doubt.
Looking at your translations, they look like machine translations with words for "boards" that are incoherent with the rest of the text, and often completely unrelated to the concept of "whiteboard".

lovasoa (changes requested):

lovasoa:
Thank you for the great PR ! Please see my comment above and fix the translations, then we can merge !

nneka-nu:
Thanks for the review. I used DeepL for translations, my mistake.
In `board.js`, Shouldn’t `localStorage.setItem` be inside the try block?

I modified the filter in `saveBoardNametoLocalStorage()` to return `name.toLowerCase() !== boardName.toLowerCase();` since `board-1` and `BOARD-1 `return the same board.

lovasoa:
No, board names are case sensitive. Just see for yourself :

https://wbo.ophir.dev/boards/board-1
https://wbo.ophir.dev/boards/BOARD-1

lovasoa:
What would be the issue with localStorage.setItem ? Now that the function is in an event handler callback, it shouldn't be an issue for it to throw.

nneka-nu:
> No, board names are case sensitive

I see. I’m using a mac, and locally the file names are case-insensitive.

I had not heard about functions not throwing in an event handler. I’ll look it up. Thanks.

lovasoa:
Functions do throw in event handlers. It's just that there is no point in catching the error, as opposed to when the throw occurs in the middle of a larger function that is called synchronously.

nneka-nu:
Got it. Thanks.

PR #186, opened 2021-05-05
Update dockerfile COPY to use node user

rothgar:
This change allows the files in /opt/app to be owned by the node user (UID 1000) so that the container can be used with or without needing to map a volume. Prior to this change if you tried to run the container it would crash because the node user didn't have permissions to write to the directory.

<sub>
</sub>


lovasoa:
Thank you @rothgar ! This was an oversight from me.

PR #193, opened 2021-05-21
Replace mover tool by selector tool

sents:
This PR replaces the mover tool (secondary for hand) with a selector tool. It enables the user to draw a box around
elements and move all the elements in the box together. Additionaly the old way of dragging elements by clicking directly on them
also works.

The PR checks if the bounding boxes of the elements are contained in the selection box. Because SVGElement.getBBox() does not include transformations I've added the necessary calculations in `client_data/js/intersect.js`. I am not sure if selection by intersecting bounding boxes would feel better but I am working to implement that too to test it out.
Eventually it would be awesome if we could define shortcuts to act on the selection for copy/paste/cut but I haven't looked into that yet.

On the server side this adds handling batch messages going client->server, which until now only went server->client.
For now it simply duplicates the `handleMessage` because I didn't want to change the server code so much.
The selector tool works well with #124.

Thank you for making this awesome tool!
<sub>
</sub>

### DIFF
--- a/client-data/tools/hand/hand.js
+++ b/client-data/tools/hand/hand.js
@@ -75,7 +75,8 @@
 	shape.width.baseVal.value = 0;
 	shape.height.baseVal.value = 0;
 	shape.setAttribute("stroke", "black");
-	shape.setAttribute("stroke-width", 3);
+	shape.setAttribute("stroke-width", 1);
+	shape.setAttribute("vector-effect", "non-scaling-stroke");
 	shape.setAttribute("fill", "none");
 	shape.setAttribute("stroke-dasharray", "5 5");
 	shape.setAttribute("opacity", 1);
--- a/client-data/tools/hand/hand.js
+++ b/client-data/tools/hand/hand.js
@@ -38,20 +38,6 @@
 	var selectorState = selectorStates.pointing;
 	var last_sent = 0;
 
-	function inRect(x, y, rect) {
-		return (x >= rect.x && x <= rect.x + rect.width) &&
-			(y >= rect.y && y >= rect.w + rect.height)
-	}
-
-	function intersectRect(rect1, rect2) {
-		return !(
-			(rect1.x + rect1.width <= rect2.x) ||
-			(rect2.x + rect2.width <= rect1.x) ||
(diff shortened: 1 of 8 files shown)


client-data/tools/hand/hand.js
```diff
@@ -50,6 +64,108 @@
...
+	tmatrix = get_translate_matrix(selectionRect);
+	tmatrix.e = 0;
+	tmatrix.f = 0;
+    }
+
+
+    function calculateSelection() {
+	var scale = Tools.drawingArea.getCTM().a;
+	var selectionTBBox = selectionRect .transformedBBox(scale);
+	return Array.from(Tools.drawingArea.children).filter(el => {
```
lovasoa, line 127: Isn't that going to just freeze the browser once there is more than just a few elements on the board ?
Did you try that on a board such as https://wbo.ophir.dev/boards/anonymous ?

lovasoa (changes requested):
Hello and thank you very much for your contribution ! Selection would be a great feature to have ! But unfortunately, you would need to find a way to let the browser's engine compute the selected elements list. Otherwise, this will not scale beyond very small boards.

client-data/tools/hand/hand.js
sents: No I haven't tried it on such a full board. Is there any way that I could download or you could send the board so I can try it?

sents: Okay, I tested it on https://wbo.ophir.dev/boards/anonymous and it handles the board without a noticeable freeze in the browser.

lovasoa: I did test it locally, and even though the framerate drops to some low values during selection, I agree that the overall performance is acceptable, and I think that the improvements are worth the cost.
So I apologize for my initial knee jerk reaction, and I'll try to review this in details and merge this week.

Thank you again.

lovasoa (approved):
Ok, I tested the changes, and made a few small changes. I am happy with how things are now. If you validate my changes, I'm ready to merge.

sents (commented):
Your changes look good to me and everything works when I test it.

lovasoa:
It's live on https://wbo.ophir.dev 👍

Issue #84, opened 2020-05-29
More control over permissions (view only / edit, maybe also access authorization?)


bessw:
It would be cool to be able to control who is allowed to edit and who is only allowed to view (and maybe use some kind of temporary marker / pointer to point out what he is talking about when asking a question).

A simple way to achieve something like this could be with to links like:
- view only:
    https://wbo.ophir.dev/boards/e2r-vvdnchZuv6gQjUJW5s/
- with edit permission (just appending an additional authorization string):
    https://wbo.ophir.dev/boards/e2r-vvdnchZuv6gQjUJW5s/NG6m4d6T0vC-GWf21GuQo-

A more secure way would be with authentication, this would also allow more fine gradient permission and access control, but at the cost of being more complex to implement.

lovasoa:
Yes, this would be cool to have :)

etsstiffler:
Would be a really nice feature! Espacially when working with younger students

PR #225, opened 2022-01-04
Add support for JWT authorisation

jamesdeacon:
These changes allow support for authorisation by JWT. 

By default it's turned off but can be set by using the two new config options `AUTH_METHOD` and `AUTH_SECRET_KEY`. The JWT should be added using the query parameter `token`.

It uses the [jsonwebtoken package](https://www.npmjs.com/package/jsonwebtoken). 

Also opens the door to solve future issues like https://github.com/lovasoa/whitebophir/issues/84 to set different access levels for different users as this could be declared in the jwt.

<sub>
</sub>

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ If you feel like contributing to this collaborative project, you can [translate 
 
 WBO supports authentication with a JWT. This should be passed in as a query with the key `token`, eg, `http://myboard.com/boards/test?token={token}`
 
-The `AUTH_SECRET_KEY` variable in [`configuration.js`](./server/configuration.js) should filled with the secret key for the JWT.
+The `AUTH_SECRET_KEY` variable in [`configuration.js`](./server/configuration.js) should be filled with the secret key for the JWT.
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ You can see a list of these variables in [`configuration.js`](./server/configura
 Some important environment variables are :
  - `WBO_HISTORY_DIR` : configures the directory where the boards are saved. Defaults to `./server-data/`.
  - `WBO_MAX_EMIT_COUNT` : the maximum number of messages that a client can send per unit of time. Increase this value if you want smoother drawings, at the expense of being susceptible to denial of service attacks if your server does not have enough processing power. By default, the units of this quantity are messages per 4 seconds, and the default value is `192`.
+ - `AUTH_SECRET_KEY` : If you would like to authenticate your boards using jwt, this declares the secret key.
 
 ## Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -79,6 +79,12 @@ See instructions on our Wiki about [how to setup a reverse proxy for WBO](https:
 WBO is available in multiple languages. The translations are stored in [`server/translations.json`](./server/translations.json). 
 If you feel like contributing to this collaborative project, you can [translate WBO into your own language](https://github.com/lovasoa/whitebophir/wiki/How-to-translate-WBO-into-your-own-language).
 
+## Authentication
+
+WBO supports authentication with a JWT. This should be passed in as a query with the key `token`, eg, `http://myboard.com/boards/test?token={token}`
+
+The `AUTH_SECRET_KEY` variable in [`configuration.js`](./server/configuration.js) should filled with the secret key for the JWT.
+
 ## Configuration
 
 When you start a WBO server, it loads its configuration from several environment variables.
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -67,7 +67,7 @@ Tools.connect = function () {
 	var token = params.get("token");
 
 	this.socket = io.connect('', {
-		"query": "token=" + token,
+		...(token && {"query" : "token=" + token}),
 		"path": window.location.pathname.split("/boards/")[0] + "/socket.io",
 		"reconnection": true,
 		"reconnectionDelay": 100, //Make the xhr connections as fast as possible
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -64,15 +64,18 @@ Tools.connect = function () {
 
 	var url = new URL(window.location);
 	var params = new URLSearchParams(url.search);
-	var token = params.get("token");
 
-	this.socket = io.connect('', {
-		...(token && {"query" : "token=" + token}),
+	var socket_params = {
 		"path": window.location.pathname.split("/boards/")[0] + "/socket.io",
 		"reconnection": true,
 		"reconnectionDelay": 100, //Make the xhr connections as fast as possible
 		"timeout": 1000 * 60 * 20 // Timeout after 20 minutes
-	});
+	}
+	if(params.has("token")) {
+		socket_params.query = "token=" + params.get("token");
+	}
+
+	this.socket = io.connect('', socket_params);
 
 	//Receive draw instructions from the server
 	this.socket.on("broadcast", function (msg) {
--- a/client-data/js/board.js
+++ b/client-data/js/board.js
@@ -62,8 +62,12 @@ Tools.connect = function () {
 		self.socket = null;
 	}
 
+	var url = new URL(window.location);
+	var params = new URLSearchParams(url.search);
+	var token = params.get("token");
 
 	this.socket = io.connect('', {
+		"query": "token=" + token,
 		"path": window.location.pathname.split("/boards/")[0] + "/socket.io",
 		"reconnection": true,
 		"reconnectionDelay": 100, //Make the xhr connections as fast as possible
--- a/server/configuration.js
+++ b/server/configuration.js
@@ -55,8 +55,6 @@ module.exports = {
   */
   STATSD_URL: process.env["STATSD_URL"],
 
-  /** Method to use for authorisation of users. Currently accepted values are 'jwt' or ''(none) */
-  AUTH_METHOD: (process.env["AUTH_METHOD"] || ""),
   /** Secret key for jwt */
   AUTH_SECRET_KEY: (process.env["AUTH_SECRET_KEY"] || ""),
 };
--- a/server/configuration.js
+++ b/server/configuration.js
@@ -54,4 +54,9 @@ module.exports = {
    * example: udp://127.0.0.1
   */
   STATSD_URL: process.env["STATSD_URL"],
+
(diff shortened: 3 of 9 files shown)


server/configuration.js
```diff
@@ -54,4 +54,9 @@ module.exports = {
    * example: udp://127.0.0.1
   */
   STATSD_URL: process.env["STATSD_URL"],
+
+  /** Method to use for authorisation of users. Currently accepted values are 'jwt' or ''(none) */
+  AUTH_METHOD: (process.env["AUTH_METHOD"] || ""),
+  /** Secret key for jwt */
+  AUTH_SECRET_KEY: (process.env["AUTH_SECRET_KEY"] || ""),
```
lovasoa, line 61: Do we need 2 environment variables ? Can we just enforce JWT verification whenever JWT_SECRET_KEY is defined and non-empty ?

lovasoa (changes requested):
Thank you for the PR, this is something that has come up from other users before, and that will be useful.

There is a comment below, plus, this needs some basic documentation for users in the readme, otherwise no one is ever going to guess that the code expects a token as a get parameter.

server/server.js
```diff
@@ -102,6 +103,23 @@ function validateBoardName(boardName) {
...
   var parts = parsedUrl.pathname.split("/");
+
+  // Check for auth if required
+  if(config.AUTH_METHOD == "jwt") {
+    if(parsedUrl.searchParams.get("token")) {
+      var token = parsedUrl.searchParams.get("token");
+      jsonwebtoken.verify(token, config.AUTH_SECRET_KEY, function(error, decoded) {
+        if(error) { // Token not valid
+          response.writeHead(403, { 'Content-Type': 'text/plain' });
+          response.end("Error: Forbidden\n" + error);
```
lovasoa, line 114: I don't think this code is correct. The verification is executed asynchronously, and does not prevent the rest of the code from executing. There is a race condition between the token verification and the happy path.

lovasoa (changes requested):

server/configuration.js
jamesdeacon: Yep, fair enough. I included both for any future alternatives, but I've reduced this to 1 in the latest commit

server/server.js
jamesdeacon: You're totally correct, I've resolved this in the latest commit

README.md
```diff
@@ -86,6 +86,7 @@ You can see a list of these variables in [`configuration.js`](./server/configura
 Some important environment variables are :
  - `WBO_HISTORY_DIR` : configures the directory where the boards are saved. Defaults to `./server-data/`.
  - `WBO_MAX_EMIT_COUNT` : the maximum number of messages that a client can send per unit of time. Increase this value if you want smoother drawings, at the expense of being susceptible to denial of service attacks if your server does not have enough processing power. By default, the units of this quantity are messages per 4 seconds, and the default value is `192`.
+ - `AUTH_SECRET_KEY` : If you would like to authenticate your boards using jwt, this declares the secret key.
```
lovasoa, line 89: You need to document how to authenticate too. How to pass the jwt to wbo ?

server/server.js
```diff
@@ -97,29 +97,33 @@ function validateBoardName(boardName) {
...
 
+/**
+ * Throws an error if the user does not have permission
+ * @param {URL} url
+ * @throws {Error}
+ */
+ function userHasPermission(url) {
+  if(config.AUTH_SECRET_KEY != "") {
+    if(url.searchParams.get("token")) {
+      var token = url.searchParams.get("token");
```
lovasoa, line 108:
```suggestion
    var token = url.searchParams.get("token");
    if(token) {
```

server/server.js
```diff
@@ -97,29 +97,33 @@ function validateBoardName(boardName) {
...
+      try {
+        jsonwebtoken.verify(token, config.AUTH_SECRET_KEY);
+      } catch(error) { // Token not valid
+        throw new Error(error)
```
lovasoa, line 112: why ?

server/server.js
```diff
@@ -113,10 +135,17 @@ function handleRequest(request, response) {
...
         // If there is no dot and no directory, parts[1] is the board name
-        boardTemplate.serve(request, response);
+        try {
+          userHasPermission(parsedUrl);
+          // User has permission so we can proceed
+          boardTemplate.serve(request, response);
+        } catch (error) {
+          response.writeHead(403, { 'Content-Type': 'text/plain' });
+          response.end("Forbidden\n" + error);
+        }
```
lovasoa, line 148: Why did you duplicate the permission check into every part of the router ?  Can we have a single check at the top ?

lovasoa (changes requested):
I am thinking : Does this handle permissions in websocket connections ? Otherwise, anyone can get access to the boards anyway over websocket.

Wouldn't it be better to pass the jwt as a cookie ?

README.md
jamesdeacon: Sure. I'll add an Authentication section to the Readme

server/server.js
jamesdeacon: Will remove

server/server.js
jamesdeacon: This is so that resources are still returned when they are requested, as a resource won't have a token attached to the request. The permission check is in every section except for when resources are requested.

server/server.js
jamesdeacon: Fixed in latest commit

README.md
```diff
@@ -79,13 +79,20 @@ See instructions on our Wiki about [how to setup a reverse proxy for WBO](https:
 WBO is available in multiple languages. The translations are stored in [`server/translations.json`](./server/translations.json). 
 If you feel like contributing to this collaborative project, you can [translate WBO into your own language](https://github.com/lovasoa/whitebophir/wiki/How-to-translate-WBO-into-your-own-language).
 
+## Authentication
+
+WBO supports authentication with a JWT. This should be passed in as a query with the key `token`, eg, `http://myboard.com/boards/test?token={token}`
+
+The `AUTH_SECRET_KEY` variable in [`configuration.js`](./server/configuration.js) should filled with the secret key for the JWT.
```
lovasoa, line 86:
```suggestion
The `AUTH_SECRET_KEY` variable in [`configuration.js`](./server/configuration.js) should be filled with the secret key for the JWT.
```

client-data/js/board.js
```diff
@@ -62,8 +62,12 @@ Tools.connect = function () {
 		self.socket = null;
 	}
 
+	var url = new URL(window.location);
+	var params = new URLSearchParams(url.search);
+	var token = params.get("token");
 
 	this.socket = io.connect('', {
+		"query": "token=" + token,
```
lovasoa, line 70: won't this add token=undefined in the (default) case where no token is defined ?

lovasoa (changes requested):
This is starting to look good :+1:  See the remarks below. Can we avoid the duplicated code in the request handler function ? Either just throw and don't catch, letting the server handle the error, or throw a specific error, and catch it specifically in a wrapper to return a clean 403.

client-data/js/board.js
jamesdeacon: Fixed in latest push

server/server.js
```diff
@@ -125,49 +150,63 @@ function handleRequest(request, response) {
...
+          .renderBoard(history_file, response)
+          .then(function () {
+            log("preview", { board: boardName, time: Date.now() - t });
+            response.end();
+          })
+          .catch(function (err) {
+            log("error", { error: err.toString(), stack: err.stack });
+            response.end("<text>Sorry, an error occured</text>");
+          });
+      } catch (error) {
```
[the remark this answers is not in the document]
jamesdeacon: Fixed in latest push

server/sockets.js
```diff
@@ -29,6 +30,20 @@ function noFail(fn) {
 
 function startIO(app) {
   io = iolib(app);
+  if (config.AUTH_SECRET_KEY) {
+    // Middleware to check for valid jwt
+    io.use(function(socket, next) {
+      if(socket.handshake.query && socket.handshake.query.token) {
+        jsonwebtoken.verify(socket.handshake.query.token, config.AUTH_SECRET_KEY, function(err, decoded) {
+          if(err) return next(new Error("Authentication error: Invalid JWT"));
+          socket.decoded = decoded;
```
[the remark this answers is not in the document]
jamesdeacon: Removed in latest push

client-data/js/board.js
```diff
@@ -67,7 +67,7 @@ Tools.connect = function () {
 	var token = params.get("token");
 
 	this.socket = io.connect('', {
-		"query": "token=" + token,
+		...(token && {"query" : "token=" + token}),
```
lovasoa, line 70: Unfortunately, we can't use "modern" js in wbo, which supports old browsers and doesn't include a js compilation step

server/server.js
```diff
@@ -122,6 +122,11 @@ function handleRequest(request, response) {
 
   if (parts[0] === "") parts.shift();
 
+  // If we're not being asked for a file, then we should check permissions.
+  if(parsedUrl.pathname.indexOf(".") === -1) {
+    checkUserPermission(parsedUrl);
```
lovasoa, line 127: It looks like you can escape permission check just by having a dot in your board name.

lovasoa (changes requested):

client-data/js/board.js
jamesdeacon: Okay, I'll put an if block around the whole lot to deal with this instead

lovasoa: I would do:

```js
        var socket_params = {
            "x": "y"
        };
        if (params.has("token")) socket_params.token = params.get("token");
	this.socket = io.connect('', socket_params)
```

jamesdeacon: 👍 Much better, will do

server/server.js
jamesdeacon: This is true, I hadn't considered that board names with a dot in would be a valid name. Perhaps I should look into including the token query with the resources too, more secure this way too. I'm unfamiliar with the templating engine but it looks straight forward enough so will look into it

jamesdeacon: I've checked and the current implementation does not support dots in board names anyway. 

I've improved the code anyway so it will now do a permissions check on any file extension that is not a js,css,svg,ico,png,jpg or gif

lovasoa (approved):

lovasoa:
I'm sorry for my lack of reactiveness ! The final version seems good to me ! Thank you again for your contribution !

Issue #212, opened 2021-06-15
Clear All button in the Menu


SushantGautam:
I think it's worth having **a clear all button** in the Menu.
We can fire an alert for confirmation to prevent accidents.
This can be disabled by default but if there is such an option in config. to enable the button it can be very handy for teachers who resue the board.

lovasoa:
Duplicate of https://github.com/lovasoa/whitebophir/issues/92

See
 - #152
 -  #202

PR #249, opened 2022-07-21
Added ability to set a moderator via JWT and added a Clear tool for moderators

jamesdeacon:
You can now add a `moderator` key within the jwt. If this is set to true then the user will have access to a Clear tool, which will wipe the entire contents of the board.

This fixes issues #92 and #212 

<sub>
</sub>


lovasoa (changes requested):
Great, thanks for the PR !

A few remarks : 

 - It looks like the server accepts "clear" requests without checking the moderator role.
 - Can we use the standard [role claim](https://www.rfc-editor.org/rfc/rfc7643#section-4.1.2) in th JWT instead of a custom "moderator" claim ?
 - Where does the SVG come from ?

jamesdeacon:
Not sure why the build test is failing. It completes all tests when I run locally and even in the test itself it is passing most of the tests, it's just the circle test that is failing, and I've not touched that code. Are you able to re-run the test?

In response to your requests
- Server checks the token before running the clear command.
- Using standard role claim now
- The SVG is from FlatIcon. I have an unlimited license to use it on websites and in software applications.

server/server.js
```diff
@@ -98,18 +98,27 @@ function validateBoardName(boardName) {
...
     var token = url.searchParams.get("token");
     if(token) {
-      jsonwebtoken.verify(token, config.AUTH_SECRET_KEY);
+      var payload = jsonwebtoken.verify(token, config.AUTH_SECRET_KEY);
+      var roles = payload.roles;
+      if(roles) {
+        return roles.includes("moderator");
+      } else {
+        return false;
+      }
```
lovasoa, line 116: Ca we avoid code duplication and use isModerator here ?

lovasoa (changes requested):
Great ! I'm ready to merge this, but you'll need to change the illustration. You need to have the right of redistributing it under [WBO's license](https://github.com/lovasoa/whitebophir/blob/master/LICENSE). 

You can create your own svg icon, or use an existing true open-source icon, such as one from [remixicon](https://remixicon.com/) for example

jamesdeacon:
I've made the code change requested and have changed the icon to one I've created

jamesdeacon:
Sorry, got confused. I've removed the duplication now. Both server.js and boardData.js now reference the new jwtauth.js file

lovasoa:
Great, thank you for your contribution!

jamesdeacon:
That's great. Thanks so much for the brilliant whiteboard! Happy to contribute

mwllgr:
Will this get published on Docker Hub, @lovasoa? Thanks for the nice feature, @jamesdeacon!

Issue #253, opened 2022-09-02
Board name in JWT


holema:
Hello,
thanks for the great tool. I realy love what you have build here.

I have a question concerning the JWT, is it possible to add the board name in the jwt. Because up to now, if I know the board name, I can enter the board with any valid key. 
I think it would be very helpfull, if the board name is also in the jwt and is protected like the role. So you have to get the correct JWT for this specific room.

Best regards

lovasoa:
That would be a great feature ! PR welcome ! But we need to take care of staying backwards-compatible.

holema:
Hello I added a PR to this feature.

holema:
#254 closes this issue

PR #254, opened 2022-09-03
add JWT board name verification

holema:
This is the PR for #253

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,11 +91,33 @@ Within the payload, you can declare the user's roles as an array. Currently the 
 {
   "iat": 1516239022,
   "exp": 1516298489,
-  "roles": ["moderator"]
+  "roles": ["moderator"],
+}
+```
+## Board name verification in the JWT
+
+WBO supports verification of the board with a JWT. The board name should be passed in the JWT in the `room_name` claim in the JWT.
+
+The `AUTH_SECRET_KEY` variable in [`configuration.js`](./server/configuration.js) should be filled with the secret key for the JWT.
+The `CHECK_BOARDNAME_IN_JWT` variable in [`configuration.js`](./server/configuration.js) should be set to `true`
+
+```
+{
+  ....
+  "board_name": "<boardName>"
+}
+```
+eg, `http://myboard.com/boards/mySecretBoarName?token={token}`
+```
+{
+  "iat": 1516239022,
+  "exp": 1516298489,
+  "roles": ["moderator"],
+  "board_name": "mySecretBoarName"
 }
 ```
 
-Moderators have access to the Clear tool, which will wipe all content from the board.
+You can now be sure that only users who have the correct token have access to the board with the specific name.
 
 ## Configuration
 
--- a/server/jwtBoardnameAuth.js
+++ b/server/jwtBoardnameAuth.js
@@ -1,0 +1,76 @@
+/**
+ *                        WHITEBOPHIR
+ *********************************************************
+ * @licstart  The following is the entire license notice for the
+ *  JavaScript code in this page.
+ *
+ * Copyright (C) 2013  Ophir LOJKINE
+ *
+ *
+ * The JavaScript code in this page is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU
+ * General Public License (GNU GPL) as published by the Free Software
+ * Foundation, either version 3 of the License, or (at your option)
+ * any later version.  The code is distributed WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU GPL for more details.
+ *
+ * As additional permission under GNU GPL version 3 section 7, you
+ * may distribute non-source (e.g., minimized or compacted) forms of
+ * that code without the copy of the GNU GPL normally required by
+ * section 4, provided you include this license notice and a URL
+ * through which recipients can access the Corresponding Source.
+ *
+ * @licend
+ */
+
+config = require("./configuration.js"),
+    jsonwebtoken = require("jsonwebtoken");
+
+/**
+ * Validates jwt and returns whether boar name fits to the boarname given in the JWT
+ * @param {URL} url
+ * @param {string} boardNameIn
+ * @returns {boolean} - True if user is a moderator, else false
+ * @throws {Error} - If no token is provided when it should be or when the board name is incorrect
+ */
+function checkBoardName(url, boardNameIn) {
+    var roomIsCorrect = true;
+    if (config.AUTH_SECRET_KEY != "") {
+        var token = url.searchParams.get("token");
+        if (token) {
+            if (config.CHECK_BOARDNAME_IN_JWT) {
+                roomIsCorrect = getBoardnamefromToken(token, boardNameIn);
+            }
+        }else {
+            throw new Error("No token provided");
+        }
+    }
+    return roomIsCorrect;
+}
+
+/**
+ * Check if user is a moderator
+ * @param {string} token
+ * @param {string} boardNameIn
+ */
+
+function getBoardnamefromToken(token, boardNameIn) {
+    if (config.AUTH_SECRET_KEY != "") {
+        var payload = jsonwebtoken.verify(token, config.AUTH_SECRET_KEY);
+        var boardName = payload.board_name;
+        if (boardName) {
+            if (boardName === boardNameIn) {
+             return true;
+            } else {
+                throw new Error("No board name match");
+            }
+        } else {
+            throw new Error("No board name provided");
+        }
+    } else {
+        return true;
+    }
+}
+
+module.exports = {checkBoardName};
--- a/server/server.js
+++ b/server/server.js
@@ -1,27 +1,27 @@
 var app = require("http").createServer(handler),
-  sockets = require("./sockets.js"),
-  {log, monitorFunction} = require("./log.js"),
-  path = require("path"),
-  fs = require("fs"),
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -91,11 +91,33 @@ Within the payload, you can declare the user's roles as an array. Currently the
 {
   "iat": 1516239022,
   "exp": 1516298489,
-  "roles": ["moderator"]
+  "roles": ["moderator"],
 }
 ```
+## Board name verification in the JWT
 
-Moderators have access to the Clear tool, which will wipe all content from the board.
```
lovasoa, line 98: Can we leave this?

README.md
```diff
@@ -91,11 +91,33 @@ Within the payload, you can declare the user's roles as an array. Currently the
...
+WBO supports verification of the board with a JWT. The board name should be passed in the JWT in the `room_name` claim in the JWT.
+
+The `AUTH_SECRET_KEY` variable in [`configuration.js`](./server/configuration.js) should be filled with the secret key for the JWT.
+The `CHECK_BOARDNAME_IN_JWT` variable in [`configuration.js`](./server/configuration.js) should be set to `true`
+
+```
+{
+  ....
+  "board_name": "<boardName>"
```
lovasoa, line 107: This is not a standard property. Can we reuse the existing roles property ?

```json
{ 
   "roles": ["editor:my_first_board", "editor:my_second_board"],
}
```

?

README.md
```diff
@@ -91,11 +91,33 @@ Within the payload, you can declare the user's roles as an array. Currently the
...
+  "roles": ["moderator"],
 }
 ```
+## Board name verification in the JWT
 
-Moderators have access to the Clear tool, which will wipe all content from the board.
+WBO supports verification of the board with a JWT. The board name should be passed in the JWT in the `room_name` claim in the JWT.
+
+The `AUTH_SECRET_KEY` variable in [`configuration.js`](./server/configuration.js) should be filled with the secret key for the JWT.
+The `CHECK_BOARDNAME_IN_JWT` variable in [`configuration.js`](./server/configuration.js) should be set to `true`
```
lovasoa, line 102: shouldn't we check it by default when it is present ?

server/server.js
```diff
@@ -1,90 +1,91 @@
...
-      MIN_NODE_VERSION +
-      " !!!"
-  );
+    console.warn(
+        "!!! You are using node " +
+        process.version +
+        ", wbo requires at least " +
+        MIN_NODE_VERSION +
+        " !!!"
+    );
```
lovasoa, line 24: Can you remove the formatting changes ? It makes reviewing more difficult

server/jwtBoardnameAuth.js
```diff
@@ -0,0 +1,76 @@
...
+ * through which recipients can access the Corresponding Source.
+ *
+ * @licend
+ */
+
+config = require("./configuration.js"),
+    jsonwebtoken = require("jsonwebtoken");
+
+/**
+ * Validates jwt and returns whether boar name fits to the boarname given in the JWT
```
lovasoa, line 31:
```suggestion
 * Validates jwt and returns whether board name fits to the board name given in the JWT
```

lovasoa (changes requested):
Thank you for the pr ! Can you remove the formatting changes ? They make it difficult to review the actual behavior changes

holema:
I changes the rooms to an object in the JWT-Claim roles.
Removed all formationg issues

README.md
```diff
@@ -94,9 +94,36 @@ Within the payload, you can declare the user's roles as an array. Currently the
...
+  "roles": {moderator:["<boardName1>","<boardName2>"],editor:["<boardName3>","<boardName4>"] }
+}
+```
+eg, `http://myboard.com/boards/mySecretBoardName?token={token}`
+```
+{
+  "iat": 1516239022,
+  "exp": 1516298489,
+  "roles": ["moderator"],
+  "roles": {moderator:["mySecretBoardName","mySecretBoardName"]}
```
lovasoa, line 120: This is not valid JSON. Can we have a list of string in roles ? See https://datatracker.ietf.org/doc/html/rfc7643#section-4.1.2

```suggestion
  "roles": ["moderator", "editor:myOtherBoardName"]
```

This way we keep backwards-compatibility, are standard-conformant, and make it easy to add roles with common JWT-compatible tools.

lovasoa (changes requested):

holema:
I added now the suggested string sollution.

server/jwtBoardnameAuth.js
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+function getBoardnamefromToken(token, boardNameIn) {
+    if (config.AUTH_SECRET_KEY != "") {
+        var payload = jsonwebtoken.verify(token, config.AUTH_SECRET_KEY);
+        var roles = payload.roles;
+        var oneHasBoardName = false;
+        var oneHasCorretBoardname = false;
+
+        var regex = new RegExp(":\ *"+boardNameIn+"$","gm");
```
lovasoa, line 63: can't the board name itself contain `:` ?

server/jwtBoardnameAuth.js
```diff
@@ -0,0 +1,91 @@
...
+        console.log(regex);
```
lovasoa, line 64: I think you forgot to remove your debugging logs

server/jwtauth.js
```diff
@@ -56,13 +56,17 @@ function checkUserPermission(url) {
       var payload = jsonwebtoken.verify(token, config.AUTH_SECRET_KEY);
       var roles = payload.roles;
       if(roles) {
-        return roles.includes("moderator");
+          for (var r in roles) {
+              var role = roles[r]
```
lovasoa, line 60: roles is supposed to be an array, you can iterate with `of` or a simple for with an index

server/jwtauth.js
```diff
@@ -56,13 +56,17 @@ function checkUserPermission(url) {
...
+              if (role.match(/moderator/gm)) {
```
lovasoa, line 61: can't the board name itself contain `moderator` ?

lovasoa (changes requested):
I think there is a problem with how the roles are parsed.

Can you add new test cases to the tests to check the behavior is correct in all cases ?

holema:
When I start the npm run test, I get the following error.

> whitebophir@1.18.0 test
> nightwatch tests && nightwatch tests --env jwt

 
An error occurred while trying to start GeckoDriver: cannot resolve path: "./node_modules/.bin/geckodriver".
Please check that the "webdriver.server_path" config property is set correctly.

   Error: Timeout while trying to connect to GeckoDriver on port 4444.
       at ClientRequest.emit (node:events:527:28)
       at Socket.socketErrorListener (node:_http_client:454:9)
       at Socket.emit (node:events:527:28)
       at emitErrorNT (node:internal/streams/destroy:157:8)
       at emitErrorCloseNT (node:internal/streams/destroy:122:3)
       at processTicksAndRejections (node:internal/process/task_queues:83:21)


Running my dev environment on windows

holema:
I just saw, there is no unittest for the JWT-Moderator function. Or did I miss something?

holema:
Added unittests

server/jwtBoardnameAuth.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+function getBoardnamefromToken(token, boardNameIn) {
+    if (config.AUTH_SECRET_KEY != "") {
+        var payload = jsonwebtoken.verify(token, config.AUTH_SECRET_KEY);
+        var roles = payload.roles;
+        var oneHasBoardName = false;
+        var oneHasCorretBoardname = false;
+        var regex = new RegExp(  "(^[^\:]*\:)("+boardNameIn+"$)","gm");
+        console.log(regex);
```
lovasoa, line 63: The board name can contain regex special characters.

Can't we first parse the roles, then work on them ?
First parse the roles to a structure like `{role: string, board_name: string}`, then work with that.
And parse with a simple 

```ts
function parse_role(role: string) {
    let [_, role_name, board_name] = role.match(/^([^:]*):?(.*)$/);
}
```

This would work for the empty string, for `:::\\`, for `"just_role"`, and for anything you can think of.
And this would make the code clearer, separating role parsing from the use of the roles.

And we don't want to leave the console logs.

lovasoa (changes requested):

server/jwtBoardnameAuth.js
```diff
@@ -0,0 +1,88 @@
...
+            throw new Error("No token provided");
+        }
+    }
+    return roomIsCorrect;
+}
+
+/**
+ * Check if user is a moderator
+ * @param {string} token
+ * @param {string} boardNameIn
```
lovasoa, line 53: Can we document what the function does and what it returns ?

server/jwtBoardnameAuth.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+function getBoardnamefromToken(token, boardNameIn) {
```
lovasoa, line 56: This function does not really "get" a board name, does it ?

I think we should have a function called 

```ts
function roleInBoard(jwt: string, board: string): "moderator" | "editor" | "forbidden" {}
```

for `roles = ["moderator"]`, it should return `moderator` for any board.


Currently you have duplicated code between getBoardnamefromToken and checkIfModerator that do not parse the role exactly the same.

holema:
All changes are done. created a new roleInBoard function. no duplicated Code anymore.

lovasoa (approved):

lovasoa:
Thank you for your contribution ! I'll release a new version

PR #255, opened 2022-09-05
Optional material design

holema:
I created the possibility to theme the board with material Design.
For this I create a new env variable: THEME_OPTIONS. this can be set to material.
Then a new css with material design is added.
Also Roboto is included as font.
image


holema:
So you have no problem when I overvrite your css with material design?
That would be great, because I think it is more up to date.

lovasoa:
The CSS should be coherent in the homepage and in the boards, but that's all. I am open to PRs that change in the CSS for the board page.

holema:
I changed the original css to material design.

client-data/board.css
```diff
@@ -1,7 +1,19 @@
...
+	font-family: 'Roboto';
+	font-style: normal;
+	font-weight: 400;
+	src: url('fonts/roboto-v30-latin-regular.eot'); /* IE9 Compat Modes */
+	src: local(''),
+	url('fonts/roboto-v30-latin-regular.eot?#iefix') format('embedded-opentype'), /* IE6-IE8 */
+	url('fonts/roboto-v30-latin-regular.woff2') format('woff2'), /* Super Modern Browsers */
+	url('fonts/roboto-v30-latin-regular.woff') format('woff'), /* Modern Browsers */
+	url('fonts/roboto-v30-latin-regular.ttf') format('truetype'), /* Safari, Android, iOS */
+	url('fonts/roboto-v30-latin-regular.svg#Roboto') format('svg'); /* Legacy iOS */
```
lovasoa, line 11: I'd like to avoid embedding large blobs in this repo.
Can we use system fonts ?

README.md
```diff
@@ -147,3 +147,11 @@ Example: `docker run -e STATSD_URL=udp://127.0.0.1:8125 lovasoa/wbo`.
...
  - If you use **prometheus**, you can collect the metrics with [statsd-exporter](https://hub.docker.com/r/prom/statsd-exporter).
  - If you use **datadog**, you can collect the metrics with [dogstatsd](https://docs.datadoghq.com/developers/dogstatsd).
+
+## Themeing
+
+it is posible to choose between different themes. Up to now only the __default__ and __material__ theme is supported.
+
+To choose the material theme add
+
+`THEME_OPTIONS=material`
```
lovasoa, line 157: Let's remove the theming options

lovasoa (changes requested):

client-data/board.css
holema: Unfortunentally Roboto is only possible as extra font.
But Arial almost looks like the same. We could also use Roboto directly from google, but this I want to avoid because then we are not independent.

But I think deliver a font is like the images we have in the repo. It`s not unusal to deliver the correct font in a software project.

README.md
holema: Sure.

client-data/board.css
holema: I just took a look. the size is only ~90kb

lovasoa: Can we just use the system's default ? On the frontend wejust use 

```css
font-family: sans-serif;
font-weight: 300;
```

which I think looks decent and well-integrated everywhere.

holema:
removed the roboto extra fonts

server/configuration.js
```diff
@@ -58,4 +58,7 @@ module.exports = {
   /** Secret key for jwt */
   AUTH_SECRET_KEY: (process.env["AUTH_SECRET_KEY"] || ""),
 
+  /** Theme options e.g. material */
+  THEME_OPTIONS: (process.env["THEME_OPTIONS"] || ""),
+
```
lovasoa, line 63: we don't need that do we ?

client-data/index.css
```diff
@@ -1,7 +1,19 @@
...
+  font-style: normal;
+  font-weight: 400;
+  src: url('fonts/roboto-v30-latin-regular.eot'); /* IE9 Compat Modes */
+  src: local(''),
+  url('fonts/roboto-v30-latin-regular.eot?#iefix') format('embedded-opentype'), /* IE6-IE8 */
+  url('fonts/roboto-v30-latin-regular.woff2') format('woff2'), /* Super Modern Browsers */
+  url('fonts/roboto-v30-latin-regular.woff') format('woff'), /* Modern Browsers */
+  url('fonts/roboto-v30-latin-regular.ttf') format('truetype'), /* Safari, Android, iOS */
+  url('fonts/roboto-v30-latin-regular.svg#Roboto') format('svg'); /* Legacy iOS */
+}
```
lovasoa, line 12: We need to remove that too don't we ?

server/server.js
```diff
@@ -108,7 +108,7 @@ function handleRequest(request, response) {
   if (parts[0] === "") parts.shift();
 
   var fileExt = path.extname(parsedUrl.pathname);
-  var staticResources = ['.js','.css', '.svg', '.ico', '.png', '.jpg', 'gif'];
+  var staticResources = ['.js','.css', '.svg', '.ico', '.png', '.jpg', '.gif', '.ttf', '.eot', '.woff', '.woff2'];
```
lovasoa, line 111: we don't have external fonts anymore

lovasoa (changes requested):

server/server.js
holema: That`s right. But in my case, I import the fonts with an own Dockerfile and replace the CSS and want to serve the Fonts via the server. So if we can leaf it with this extensions, I would be very happy.

client-data/index.css
holema: ✔️

server/configuration.js
holema: ✔️

holema:
Hello @lovasoa,

will you merge the PR in the future?

Best regards

holema:
Hello @lovasoa,
Will you merge the PR?

PR #256, opened 2022-09-08
FIX - Server checks token to see if user is moderator before clear

andrea78:
Thank you for a new useful function.

<sub>
</sub>

### DIFF
--- a/server/boardData.js
+++ b/server/boardData.js
@@ -30,7 +30,7 @@ var fs = require("./fs_promises.js"),
   path = require("path"),
   config = require("./configuration.js"),
   Mutex = require("async-mutex").Mutex,
-  jwtauth = require("./jwtauth.js");
+  jwtBoardnameAuth = require("./jwtBoardnameAuth.js");
 
 /**
  * Represents a board.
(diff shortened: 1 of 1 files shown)


server/boardData.js
```diff
@@ -174,7 +174,7 @@ class BoardData {
         this.addChild(message.parent, message);
         break;
       case "clear":
-        if(jwtauth.checkIfModerator(message.token)) {
+        if(jwtBoardnameAuth.roleInBoard(message.token) === "moderator") {
```
lovasoa, line 177: Shouldn't we check that the user is a moderator *in the current board* ? We now have roles like `moderator:board77`

lovasoa (changes requested):

server/boardData.js
andrea78: It depends on the concept. If we are going to be backwards compatible, I don't think it is needed as all moderators have access to the Clear tool, which will remove all content from the board. For example, if moderators do not have a board assigned to them, they can remove all content from each board. On the other hand, if the moderators have a board assigned to them, they can only remove a content from the board assigned to them, because they do not have the right to access the boards of other moderators who have the board assigned to them.

holema: The Function returns the role moderator, when The board name is correct with the board name in the url, and returns moderator when there is set the role moderator in the roles claim but no boardname is set. in this case a user with such a JWT can wipe every board.

lovasoa: Yes, you have to give the board name to the roleInBoard function, and it will do the right thing if the role is set to just `"moderator"`.

PR #274, opened 2023-04-25
Add viewer privilage

rohith1222004:
I've made an update to our whiteboard application that adds a new "viewer" permission to the JWT payload. This update will allow us to restrict users to only viewing the whiteboard, without giving them editing capabilities.

### DIFF
--- a/index.py
+++ b/index.py
@@ -1,0 +1,11 @@
+# encode.py
+import datetime
+import jwt # import jwt library
+SECRET_KEY = "test123"
+# json data to encode
+json_data = {
+  "roles": ["moderator:firstboard","viewer:hboard"]
+}
+encode_data = jwt.encode(payload=json_data, \
+                        key=SECRET_KEY, algorithm="HS256")
+print(encode_data) 
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "jsonwebtoken": "^8.5.1",
     "polyfill-library": "^3.107.1",
     "serve-static": "^1.14.1",
-    "socket.io": "^4",
+    "socket.io": "^4.6.1",
     "statsd-client": "^0.4.7"
   },
   "scripts": {
--- a/server/templating.js
+++ b/server/templating.js
@@ -70,14 +70,26 @@ class Template {
   }
 }
 
+
+const jwtBoardnameAuth = require("./jwtBoardnameAuth"); // have to delete
 class BoardTemplate extends Template {
   parameters(parsedUrl, request, isModerator) {
     const params = super.parameters(parsedUrl, request, isModerator);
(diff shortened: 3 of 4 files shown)


index.py
```diff
@@ -0,0 +1,11 @@
...
+import datetime
+import jwt # import jwt library
+SECRET_KEY = "test123"
+# json data to encode
+json_data = {
+  "roles": ["moderator:firstboard","viewer:hboard"]
+}
+encode_data = jwt.encode(payload=json_data, \
+                        key=SECRET_KEY, algorithm="HS256")
+print(encode_data)
```
lovasoa, line 11:
```suggestion
```

We probably don't want to create a python script here

package.json
```diff
@@ -14,7 +14,7 @@
     "jsonwebtoken": "^8.5.1",
     "polyfill-library": "^3.107.1",
     "serve-static": "^1.14.1",
-    "socket.io": "^4",
+    "socket.io": "^4.6.1",
```
lovasoa, line 17: This does not seem to be linked to the current PR. You can open a distinct pr to update dependencies

server/templating.js
```diff
@@ -70,14 +70,26 @@ class Template {
   }
 }
 
+
+const jwtBoardnameAuth = require("./jwtBoardnameAuth"); // have to delete
 class BoardTemplate extends Template {
   parameters(parsedUrl, request, isModerator) {
     const params = super.parameters(parsedUrl, request, isModerator);
     const parts = parsedUrl.pathname.split("boards/", 2);
+    console.log(parts[1]);
```
lovasoa, line 79:
```suggestion
```

server/templating.js
```diff
@@ -70,14 +70,26 @@ class Template {
...
     params["board"] = decodeURIComponent(boardUriComponent);
-    params["hideMenu"] = parsedUrl.query.hideMenu == "true" || false;
+    const query = parsedUrl.query;
+    const token = query.token;
+    const boardName = parts[1];
+    const userRole = jwtBoardnameAuth.roleInBoard(token, boardName);
+    if (userRole === "viewer") {
+      params["hideMenu"] = true;
+    } else {
+      params["hideMenu"] = false;
```
lovasoa, line 90:
```suggestion
      params["hideMenu"] = false;
```

you don't want to override that. keep the value from the query

lovasoa (changes requested):
Thank you for opening a PR ! 

However, it looks like the new implementation does not actually prevent the user from sending write commands to the socket.io handler. Users can actually write to the board even with a simple viewer token, the menu is just hidden in the ui

lovasoa:
I've converted the PR to a draft, feel free to pass it back to "ready for review" when you want me to look at it again

PR #285, opened 2023-10-25
Sandstorm app package support

### LINKED ISSUE #284: Note: There is now a Sandstorm package for this app!
Hello, I thought you might like to know I recently packaged WBO for Sandstorm. I need to finish filling out some of the metadata, but the experimental release is here: https://apps.sandstorm.io/app/rhtntjye2n7p71vrcq0ae4f6n9nr139fdk4n884tw55k40vy4uph?experimental=true (and the final one will be without the ?experimental=true once published).

Currently it is a very simple fork: As on Sandstorm, apps should preferably be single-document, the index.html has been replaced with a meta-refresh to a board named "board", and I had to clear out a call for os.userInfo() which fails in the Sandstorm environment.

ocdtrekkie:
With these changes, the Sandstorm packaged version can run the same code as a standard deployment with a single additional configuration option. I am not much of a JavaScript developer, so nitpicks are welcome.

Fixes #284 
<sub>
</sub>


.sandstorm/Vagrantfile
```diff
@@ -0,0 +1,111 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# CAUTION: DO NOT MAKE CHANGES TO THIS FILE. The vagrant-spk upgradevm process will overwrite it.
```
lovasoa, line 4: is this file auto-generated ? If so, then maybe it shouldn't be on git ?

.sandstorm/pgp-signature
lovasoa, line 1: do these file need to be on git ?

server/check_output_directory.js
```diff
@@ -16,7 +16,12 @@ async function get_error(directory) {
   if (!fs.statSync(directory).isDirectory()) {
     error = "exists, but is not a directory";
   }
-  const { uid, gid } = os.userInfo();
+  try {
+    const { uid, gid } = os.userInfo();
+  } catch (e) {
+    console.error(`os.userInfo does not work in this environment.`);
+    const { uid, gid } = [];
+  }
```
lovasoa, line 24: this does not work, does it ? uid and gid become local to the scope of the try/catch.

Looking at the code, we could avoid the call to `userInfo` completely in the normal case, since we use the uid/gid only when the writeFileSync fails.

```js
  try {
    fs.writeFileSync(tmpfile, "{}");
    fs.unlinkSync(tmpfile);
  } catch (e) {
    let err_msg = "does not allow file creation and deletion. ";
    try {
        const { uid, gid } = os.userInfo();
        err_msg += "Check the permissions of the directory, and if needed change them so that " +
      `user with UID ${uid} has access to them. This can be achieved by running the command: chown ${uid}:${gid} on the directory`;
      } finally {
             return err_msg;
      }
  }
```

server/configuration.js
```diff
@@ -57,5 +57,8 @@ module.exports = {
 
   /** Secret key for jwt */
   AUTH_SECRET_KEY: (process.env["AUTH_SECRET_KEY"] || ""),
+  
+  /** If this variable is set, automatically redirect to this board. */
```
lovasoa, line 61:
```suggestion
  /** If this variable is set, automatically redirect to this board from the root of the application. */
```

server/server.js
```diff
@@ -223,7 +223,11 @@ function handleRequest(request, response) {
 
     case "": // Index page
       logRequest(request);
+      if (!config.DEFAULT_BOARD)
         indexTemplate.serve(request, response);
+      else
```
lovasoa, line 228: the double negation is hard to process. Maybe invert the if and else: `if (config.DEFAULT_BOARD) ... else ...`

server/server.js
```diff
@@ -223,7 +223,11 @@ function handleRequest(request, response) {
...
+        response.writeHead(302, { Location: 'boards/' + config.DEFAULT_BOARD });
```
lovasoa, line 229: The board name needs to be urlencoded.

lovasoa (changes requested):
Thanks !

I don't have the competence to review and maintain the things in `.sandstorm`. I'll merge the changes to the application itself though, after you fix my little nitpicks :)

ocdtrekkie:
These changes worked for me in the sense that the Sandstorm package still works, and without the environment variable set, the index loads, but I suspect you are correct that I never actually test a path where the userInfo is actually used, and your solution is better. I will amend my changes tonight, along with the other changes.

As far as the contents of .sandstorm, generally, yes they all are committed together. pgp-signature is actually a file specific to this package which states I control the app publishing key for this package. While we discourage the Vagrantfile and global-setup.sh being edited, they carry some important information to replicate the build environment, such as what box is used, and regenerating them is only done in breaking scenarios, usually.

Honestly one of my unsolved problems is I did notice you don't commit your screenshots to the repo, and for the Sandstorm app metadata, we generally would. It is possible that upstreaming the package metadata is the wrong call if you prefer to keep a fairly lean tree: I potentially could PR the changes to WBO itself, but keep the .sandstorm directory on its own repository.

Instead of it being a fork of WBO, I can have the separate build repo clone the official WBO repo in the build process, which could be used verbatim since I'm upstreaming the needed changes to WBO to work on Sandstorm.

lovasoa:
> I potentially could PR the changes to WBO itself, but keep the .sandstorm directory on its own repository.
> Instead of it being a fork of WBO, I can have the separate build repo clone the official WBO repo in the build process, which could be used verbatim since I'm upstreaming the needed changes to WBO to work on Sandstorm.


I think that would work better for me, yes !

ocdtrekkie:
Awesome. I will do some reorganizing on my end and submit a PR with just the improved changes to your code.

PR #287, opened 2023-11-18
Image Uploads

atomdmac:
# Description

Drag images (JPG, GIF or PNG) onto the canvas to add them to the whiteboard.  Images are saved to disk into a directory with the name name as the board JSON file. 

## Notable Code Changes

* Use new class `BoardDataList` instead of plain JS object for `boards` in `sockets` module.
  * Provides access to the boards list via functions which allows for better access control and observability.  This is desirable now that multiple objects may attempt to mutate the data.
* Add new dependency `multiparty` for parsing `FormData` on server.

## TODO
* [x] Delete images when removed from the board
* [x] Support clicking on canvas with image tool to add image
* [x] Delete tmp file after uploading image
* [x] Address possible security issue [here](https://github.com/lovasoa/whitebophir/pull/287#discussion_r1398260185)
* [x] When serving board image assets, don't block handler function

## Future Improvements
* [ ] Display upload progress indicator for image
* [ ] Support for "Download Image"


lovasoa:
Thank you very much for contributing this long-requested feature !

I'll try to review it soon, but before:
 - can we add tests ? This is a sensitive feature with big security implications, I think it needs tests.
 - the image upload endpoint needs to be disabled when the image tool is disabled (using `BLOCKED_TOOLS`)

lovasoa:
I see there are also still TODOs in the code. Feel free to switch it back to ready for review when you are ready for me to have a look

server/server.js
```diff
@@ -157,7 +167,75 @@ function handleRequest(request, response) {
...
+        response.writeHead(200, {
+          "Content-Type": "application/json",
+          "Content-Security-Policy": CSP,
+          "Cache-Control": "public, max-age=30",
+        });
+      response.end(JSON.stringify({ status: 'ok' }));
+      break;
+    case "board-assets":
+        const [, boardId, assetName] = parts;
+        const file = fs.readFileSync(path.join(config.HISTORY_DIR, `board-${boardId}`, assetName));
```
lovasoa, line 223: also, we cannot have blocking code like this in the request handling function

client-data/tools/image/image.js
```diff
@@ -0,0 +1,276 @@
...
+  var svg = Tools.svg;
+  /**
+    * Creates a new image element on the canvas, or updates an existing image
+    * with new information.
+    * @param {Object} data - The data to use to create the image.
+    */
+  function createImageElement(data) {
+    var img = svg.getElementById(data.id) || Tools.createSVGElement("image");
+    img.setAttribute("id", data.id);
+    img.setAttribute("href", data.src);
```
lovasoa, line 83: This is a big security vulnerability, isn't it ?

lovasoa: Anyone can send an image event with the `src` attribute they want

atomdmac: Is the concern that a user of Board A would be able to load images from Board B (which may be a "private" board)?

server/server.js
atomdmac: I can do some more research on this but do you happen to know if calling a function that returns a promise would solve this issue?

Something like this:

```js
async function readFile(request, response) {
  // ... create promise, read file from disk, and handle response.
}

function handleRequest(request, response) {
  // ... handlers
  case 'board-assets':
     serveFile(request, response);
     break;
  // ... other handlers
}
```

client-data/tools/image/image.js
lovasoa: The bigger concern is that they may have all users make requests to their own server

atomdmac: Ah OK - so the concern is that an attacker could inject a URL that points to a server that they control, and that serves files with malicious content.  Is that right?

lovasoa: Yes, they could serve whatever they want, but also log the ip addresses of everyone connected

server/server.js
lovasoa: you can look at how it's done in the rest on the code; we use async fs functions

client-data/tools/image/image.js
atomdmac: I believe this has been addressed with the latest commit 👍 Let me know if you disagree.

client-data/tools/image/image.js
```diff
@@ -80,7 +89,7 @@
   function createImageElement(data) {
     var img = svg.getElementById(data.id) || Tools.createSVGElement("image");
     img.setAttribute("id", data.id);
-    img.setAttribute("href", data.src);
+    img.setAttribute("href", getAbsoluteImageUrl(data.src));
```
lovasoa, line 92: this is complicated. Can we remove the src altogether ?

```suggestion
    img.setAttribute("href", "./images/" + data.id);
```

that would stress me less :)

atomdmac: How would the browser know where to pull the image to display from?

lovasoa: The browser would make a request to `boards/{boardname}/images/{image_id}`, the server would check that the image exists, and if so, serve it.

atomdmac: Yours is a much better approach 😅 This has been updated.

atomdmac:
Progress update: I've added some tests but for some reason I can't get them to pass in FireFox (though they pass fine in Chromium).

PR #288, opened 2023-11-20
Save user preferences

atomdmac:
Saves and restores various bits of application state:
* Stroke size
* Stroke opacity
* Stroke color
* Grid visibility and style
* Selected tool

As part of this work, a reusable "Signals" facility has been added via `createSignal`.  This allows multiple functions to subscribe to state changes.

<sub>
</sub>


lovasoa:
Hey, thanks for the PR! I'm not sure what the use case for this is, though; and I don't think I'd like the brush parameters to be saved. If you have two windows opened, they will be constantly fighting for control over the saved parameters.

atomdmac:
I actually found it frustrating that I'd lost my settings between refreshes.  What if the settings were saved per boars instead of globally?

Issue #317, opened 2024-12-13
Create an exit button that closes the window


RHJasonSmith:
I'm running wbo in kiosk mode on a pi, and will occasionally need to exit the window to do something on the device.

PR #318, opened 2024-12-13
Add exit button

SiliconValet:
https://github.com/lovasoa/whitebophir/issues/317

Adds an exit button with configuration to enable.

<sub>
</sub>


lovasoa (changes requested):
This seems very specific, and not really useful outside of your case. Maybe this should be an external plugin instead, and what we should have in the code is something like `WBO_EXTERNAL_TOOLS` that you would set to a value like `["http://example.com/yourtool.js"]` ?

SiliconValet:
I didn't see a way to easily conditionally inject the tool js without resorting to a build process  - If you've got a rough direction I can see if I can find time to whip it up.

SiliconValet:
Note that the exit button only shows up if the env var or config is set, invisible and missing by default.

PR #320, opened 2025-01-07
Update GitHub Actions to support GitHub Container Registry

### LINKED ISSUE #319: Add docker image to ghcr
Hi, any chance to push also the image to ghcr.io here? This will reduce the hub docker pull limitation.

webysther:
Closes #319 

<sub>
</sub>

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -38,7 +38,7 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.DOCKER_PASSWORD }}
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v2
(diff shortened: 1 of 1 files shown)


lovasoa:
hmm, I'm not sure, but I don't think this is how it is supposed to work? 

You are never supposed to create GitHub tokens for gh actions, they should use their own temporary token, which is more secure. Shouldn't they?

webysther:
> hmm, I'm not sure, but I don't think this is how it is supposed to work?
> 
> You are never supposed to create GitHub tokens for gh actions, they should use their own temporary token, which is more secure. Shouldn't they?

Yeap, I forgot, the `GITHUB_TOKEN` you don't need to create.

PR #335, opened 2026-04-10
[codex] Add read-only boards

lovasoa:
## Summary

This PR adds persisted read-only boards and enforces write permissions server-side.

## What changed

- store board metadata in the board JSON using `__wbo_meta__.readonly`
- block writes on read-only boards unless the connected user has `editor` or `moderator` access to that board
- block all writes on read-only boards when JWT auth is disabled
- keep board metadata out of normal board rendering and SVG export paths
- expose board state to the client and hide editing tools for read-only viewers
- update the README to clearly explain roles, board visibility, and how to switch a board between writable and read-only
- extend integration coverage for read-only behavior in authenticated and unauthenticated setups

## Why

Before this change, WBO had board access control and moderator/editor roles, but no way to make a board viewable without allowing arbitrary writes. This adds a board-level editability setting without introducing new roles.

## Impact

- operators can mark individual boards read-only by editing the persisted board file
- viewers can still access a read-only board when their token allows access
- only editors and moderators can modify a read-only board
- moderators still keep exclusive access to Clear

## Validation

- `npm test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,13 +116,13 @@ If `AUTH_SECRET_KEY` is set, opening a board requires a valid token. You can the
 
 ```json
 {
-  "roles": ["editor:board-a", "moderator:board-b", "access:board-c"]
+  "roles": ["editor:board-a", "moderator:board-b", "reader:board-c"]
 }
 ```
 
 - `editor:<boardName>` allows editing that board.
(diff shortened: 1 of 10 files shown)


lovasoa:
this supersedes https://github.com/lovasoa/whitebophir/pull/274

Issue #324, opened 2025-06-15
I am unable to find the "Select All" option.


legkamran:
Is there any option similar to "Select All"? Something like Ctrl + A?
It is possible to select a portion using the "Select" tool, but there seems to be no shortcut for deleting (although the lack of keyboard shortcuts is not a major issue, as the left toolbar is well designed — still, having them would be helpful). Currently, it is necessary to click the "X" button.

I could not find a "Select All" option to clear the entire board. Having a dedicated "Clear Board" function is not essential if there is an alternative such as "Select All" combined with a delete option.

Another issue arises when the selected area does not fit within the visible screen — in such cases, the "Copy" and "Delete" buttons do not appear (I am referring to situations where the selected area is larger than the visible portion of the screen, particularly when the top-left of the selection is outside the visible area).
It would be more convenient if these buttons remained visible at all times in a fixed and accessible location, such as under the Opacity setting or a similar section.

**Thank you very much for your work! It is truly an excellent, simple, fast, and user-friendly whiteboard system!**

legkamran:
**Edit**: Since I use Vimium C, I initially didn’t realize that the keyboard shortcuts were working.

**Issue**: If the "Select" is activated first (by double-pressing H at the beginning), followed by switching to the "Pen" (using the P key), and then pressing H again on the keyboard, the previously selected "Select" becomes active again (when switching to another tool "Select" does not revert back to the "Hand" as its default which is great). However, in this state, pressing the Delete key does not delete the selected area. If, after choosing the "Pen", I switch to the "Hand" and then back to "Select", the selected area can then be deleted using the Delete key.

PR #416, opened 2026-07-09
feat: add select-all shortcut for selector

11suixing11:
## Summary
- add `Ctrl/Cmd + A` support for the hand tool selector mode
- reuse the existing selection chrome so full-board selections can be deleted immediately
- add a regression test covering select-all followed by delete

## Why
`lovasoa/whitebophir#324` asks for a way to select everything on the board instead of relying on the floating selection controls staying visible.

This keeps the change narrow: it only applies while the selector sub-tool is active, and it leaves the existing box-selection path untouched.

## Testing
- `node --test test-node/client_tool_replay.test.js`
- `npx @biomejs/biome check client-data/tools/hand/index.js test-node/client_tool_replay.test.js`

### DIFF
--- a/client-data/tools/hand/index.js
+++ b/client-data/tools/hand/index.js
@@ -59,7 +59,7 @@ import { TOOL_CODE_BY_ID } from "../tool-order.js";
 /** @typedef {HandPointSelection | HandScaleSelection | HandPanSelection | null} HandSelection */
 /** @typedef {{x: number, y: number} | {a: number, d: number, e: number, f: number} | undefined} SelectionRectTransform */
 /** @typedef {(x: number, y: number, force: boolean) => void} HandTransformHandler */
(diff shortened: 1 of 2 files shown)


lovasoa (changes requested):
Full-board move/delete is explicitly disabled to prevent abuse. You can erase the entire board when you are a moderator. You can become a moderator either by using jwt auth, or the `WBO_BOARD_MODERATORS='boardname:usersecret'` env var: https://github.com/lovasoa/whitebophir/blob/master/server/configuration.mjs#L125-L128

11suixing11:
Thanks for explaining the abuse-prevention boundary. I missed that full-board move/delete is intentionally restricted to moderators, so this shortcut is not appropriate for the project security model. Closing this PR rather than trying to work around that design.

PR #422, opened 2026-09-08
Secure dockerfile and compose setup

gigirassy:
I love this program, but I noticed that the official Docker setup has some security pitfalls. Since security clearly matters to the maintainers (see #169 ), I contributed the Dockerfile setup I use on my instance. However, the changes are breaking to existing setups, so I understand if you're hesitant.
Here's a summary of the changes I made:
1. Multistage build to reduce image size significantly, reducing pull times.
2. I noticed that the existing Dockerfile uses set-cap to allow node to bind to a privileged port. Unfortunately, this makes the container less secure as it is not possible to drop privileges in a compose setup due to the cap set to node's binary. I went ahead and changed it so it sets the port to 8000, which is not privileged.
3. Instead of having a runtime with a shell, I went with Chainguard's reliable distroless Node runtime, which has no shell and runs as user ``node``, UID:GID 65532:65532. This means existing users gotta rechown to what Chainguard establishes, and also change the local port in their compose files. Again, I understand if you're hesitant.
4. Reconfigured the compose.yml to be read-only and drop all privileges/caps. This is a setup I've been using for a couple weeks now, and everything works as it should.
<sub>
</sub>


lovasoa (changes requested):
Hello and thank you very much for your contribution ! 

Changing the default listening port would be a breaking change, I'd rather avoid that.
I do not think any of what you report actually qualifies as a security vulnerability, but the rest of the hardening is welcome !

If you do find vulnerabilities in the future, you can report them in https://github.com/lovasoa/whitebophir/security

Thanks again for looking into this and for helping improve the project!

